PR #15, opened 2017-05-24
Add start time and end time to subscriptions

parth21shah:
Add get/set methods for start and end time in subscription object

cc @nickyr


src/Message/FetchSubscriptionRequest.php
```diff
@@ -131,6 +133,8 @@ public function getData()
             $data['vid'] = $this->getSubscriptionReference();
         }
 
+        $data['startTime'] = $startTime;
+        $data['endTime'] = $endTime;
```
nickyr: the changes in this file are unnecessary... they would add startTime and endTime parameters to the FetchSubscriptionRequest, but that API call doesn't use them.

PR #16, opened 2017-05-24
Add a status to subscription object

parth21shah:
cc @nickyr

### DIFF
--- a/src/TestFramework/DataFaker.php
+++ b/src/TestFramework/DataFaker.php
@@ -202,7 +202,7 @@ class DataFaker
      */
     public function subscriptionStatus()
     {
-        $statuses = array("Active", "Suspended", "Cancelled");
+        $statuses = array('Active', 'Suspended', 'Cancelled');
         shuffle($statuses);
         return current($statuses);
     }
--- a/src/TestFramework/DataFaker.php
+++ b/src/TestFramework/DataFaker.php
@@ -195,6 +195,18 @@ class DataFaker
         return $result;
     }
 
+     /**
+     * Return a subscription status
+     *
+     * @return string
+     */
+    public function subscriptionStatus()
+    {
+        $statuses = array("Active", "Suspended", "Cancelled");
+        shuffle($statuses);
+        return current($statuses);
+    }
+
     /**
      * Return a product id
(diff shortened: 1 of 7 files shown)


src/TestFramework/DataFaker.php
```diff
@@ -195,6 +195,18 @@ public function subscriptionReference()
...
     }
 
+     /**
+     * Return a subscription status
+     *
+     * @return string
+     */
+    public function subscriptionStatus()
+    {
+        $statuses = array("Active", "Suspended", "Cancelled");
```
nickyr: single quotes

src/TestFramework/DataFaker.php
```diff
@@ -195,6 +195,18 @@ public function subscriptionReference()
...
+        shuffle($statuses);
```
nickyr: instead of using shuffle and current, you could do

```php
return $statuses[$this->intBetween(0, count($statuses) - 1)];
```

That way, we can be sure it's using the same seed as everything else in the DataFaker. This is helpful because then if you want to debug something, you can just set an explicit seed in the DataFaker constructor and then every time the tests run they'll use the same values.

tests/Message/FetchSubscriptionRequestTest.php
```diff
@@ -38,6 +38,7 @@ public function setUp()
         $this->ipAddress = $this->faker->ipAddress();
         $this->startTime = $this->faker->timestamp();
         $this->endTime = $this->faker->timestamp();
+        $this->subscription_status = $this->faker->subscriptionStatus();
```
nickyr: camel case (that's Omnipay's style guidelines)

tests/SubscriptionTest.php
```diff
@@ -224,6 +224,17 @@ public function testEndTime()
...
      * @return void
      */
+    public function testStatus()
+    {
+        $status = $this->faker->subscriptionStatus();
+        $this->assertSame($this->subscription, $this->subscription->setStatus($status));
+        $this->assertSame($status, $this->subscription->getStatus());
+    }
+
+
```
nickyr: extra newline

PR #17, opened 2017-05-30
Add billing day to Subscription object

parth21shah:
cc @nickyr

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -489,7 +489,7 @@ class Subscription
     }
 
     /**
-     * Get the billing day of the subscription.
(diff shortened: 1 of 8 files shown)


src/Subscription.php
```diff
@@ -489,6 +489,27 @@ public function setStatus($value)
     }
 
     /**
+     * Get the billing day of the subscription.
```
nickyr: an explanation of what this is would be useful

parth21shah:
I do have it https://github.com/vimeo/omnipay-vindicia/blob/master/tests/Message/CreateSubscriptionRequestTest.php#L394 
and I also added a test in SubscriptionTest.php with this PR

PR #18, opened 2017-06-05
Add product index to create subscription request

parth21shah:
cc @nickyr


parth21shah:
We only have the product reference if we are fetching product from vindicia otherwise we do not store that on our end so, it's not possible to pass that as parameter

PR #19, opened 2017-06-05
Add subscription billing state

parth21shah:
Subscription billing state and status will help decide subscription's billing status. For example subscription status is active and billing state is "In Retry" it means subscription is active but in grace period.

cc @nickyr

### DIFF
--- a/src/Message/Response.php
+++ b/src/Message/Response.php
@@ -339,8 +339,9 @@ class Response extends AbstractResponse
      */
     public function getSubscriptionStatus()
     {
-        if (isset($this->data->autobill)) {
-            return $this->data->autobill->status;
+        $subscription = $this->getSubscription();
+        if ($subscription) {
+            return $subscription->getStatus();
         }
         return null;
     }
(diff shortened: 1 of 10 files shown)


src/TestFramework/DataFaker.php
```diff
@@ -208,6 +208,25 @@ public function subscriptionStatus()
...
+    {
+        $billingState = array(
+            'Unbilled',
+            'Good Standing',
+            'Free/Trial',
+            'In Retry',
+            'Unusable Payment Method',
+            'Billing Completed'
+        );
+        shuffle($billingState);
```
nickyr: instead of doing `shuffle`, can you use the $this->intBetween function to choose a random position? That way everything uses the number generator that was seeded in the constructor and you can just set the seed in one place to get consistent results every time

nickyr: you could do that for subscriptionStatus above also

src/Message/Response.php
```diff
@@ -346,6 +346,19 @@ public function getSubscriptionStatus()
     }
 
     /**
+     * Get the billing state of the subscription.
+     *
+     * @return string|null
+     */
+    public function getSubscriptionBillingState()
+    {
+        if (isset($this->data->autobill)) {
```
nickyr: you can just do $this->getSubscription() instead of $this->data->autobill. See getSubscriptionId above

PR #24, opened 2017-07-17
Deprecate refund 'note' in favor of 'reason'

nickyr:
Because 'reason' makes more sense.

### DIFF
--- a/src/TestFramework/DataFaker.php
+++ b/src/TestFramework/DataFaker.php
@@ -1007,6 +1007,16 @@ class DataFaker
     }
 
     /**
+     * Return a refund reason
+     *
+     * @return string
+     */
+    public function refundReason()
+    {
+        return $this->note();
+    }
+
+    /**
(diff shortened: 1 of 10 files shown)


src/TestFramework/DataFaker.php
```diff
@@ -1007,6 +1007,16 @@ public function soapId()
     }
 
     /**
+     * Return a refund reason
+     *
+     * @return string
+     */
+    public function refundReason()
+    {
+        return $this->note();
```
parth21shah: Should this function be renamed?

nickyr: @parth21shah since it's in the data faker, i just added a new function for refundReason and left `note` since something else could have a note maybe. If you prefer I could mark the note function as deprecated.

parth21shah: No it's fine!!

PR #33, opened 2018-06-07
Added page and page size fields to fetch transactions request class

rondogency:
Added page and page size fields to fetchTransactionsRequest message class

### DIFF
--- a/src/Message/FetchTransactionsRequest.php
+++ b/src/Message/FetchTransactionsRequest.php
@@ -67,6 +67,8 @@ use Omnipay\Common\Exception\InvalidRequestException;
  */
 class FetchTransactionsRequest extends AbstractRequest
 {
+    const DEFAULT_PAGE_SIZE = 10000;
+
     /**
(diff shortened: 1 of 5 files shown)


src/Message/FetchTransactionsRequest.php
```diff
@@ -67,6 +67,8 @@
  */
 class FetchTransactionsRequest extends AbstractRequest
 {
+    const DEFAULT_PAGE_SIZE = 10000;
```
parth21shah: I think it should set to 1000 at max otherwise it will surely fails

src/Message/FetchTransactionsRequest.php
```diff
@@ -53,10 +53,12 @@
  *       var_dump($fetchResponse->getTransactions());
  *   }
  *
- *   // alternatively, you could fetch by a time range
+ *   // alternatively, you could fetch by a time range and optionally page and page size
```
nickyr: can you mention page and pageSize in the bulleted list above as well?

PR #35, opened 2018-06-14
Fix PayPal email field from Vindicia response typo

rondogency:
Vindicia misspells 'payPalEmail' to be 'paypalEmail' in their fetch transaction response, which causes trouble for us to update payment method in our database.


src/ObjectHelper.php
```diff
@@ -50,6 +50,16 @@ public function buildTransaction(stdClass $object)
             }
         }
 
+        // Vindicia screw up the paypal email field by a camel case typo in the response
+        // we need to deal with "payPalEmail" and "paypalEmail" in case they fix it on their side
+        $paypal_email = null;
```
nickyr: should be $paypalEmail

src/ObjectHelper.php
```diff
@@ -50,6 +50,16 @@ public function buildTransaction(stdClass $object)
...
+        if (isset($object->statusLog[0]->payPalStatus->payPalEmail)) {
+            $paypal_email = $object->statusLog[0]->payPalStatus->payPalEmail;
+        }
+        if (isset($object->statusLog[0]->payPalStatus->paypalEmail)) {
```
nickyr: this could be an `else if`

tests/Message/FetchTransactionRequestTest.php
```diff
@@ -253,4 +255,107 @@ public function testSendByReferenceFailure()
 
         $this->assertNull($response->getTransaction());
     }
+
+    /**
+     * @return void
+     */
+    public function testSendPayPalSuccess()
+    {
+        $this->setMockSoapResponse('FetchTransactionPayPalSuccess.xml', array(
```
nickyr: can you name this FetchPayPalTransactionSuccess? That's the way other things are named

tests/Mock/FetchTransactionPayPalFailure.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nickyr: this is the same as FetchTransactionFailure.xml, so we don't need to make this a separate test

tests/Mock/FetchTransactionPayPalSuccess.xml
```diff
@@ -0,0 +1,116 @@
...
+          <sortOrder xmlns="" xsi:type="xsd:int">1</sortOrder>
+          <active xmlns="" xsi:type="xsd:boolean">1</active>
+        </sourcePaymentMethod>
+        <statusLog xmlns="" xsi:type="vin:TransactionStatus">
+          <status xmlns="" xsi:type="vin:TransactionStatusType">Authorized</status>
+          <timestamp xmlns="" xsi:type="xsd:dateTime">2018-06-15T08:06:05-07:00</timestamp>
+          <paymentMethodType xmlns="" xsi:type="vin:PaymentMethodType">PayPal</paymentMethodType>
+          <payPalStatus xmlns="" xsi:type="vin:TransactionStatusPayPal">
+            <token xmlns="" xsi:type="xsd:string">[TOKEN]</token>
+            <paypalEmail xmlns="" xsi:type="xsd:string">[PAYPAL_EMAIL]</paypalEmail>
```
nickyr: This is the spelling they actually return, right?

src/ObjectHelper.php
```diff
@@ -50,6 +50,15 @@ public function buildTransaction(stdClass $object)
             }
         }
 
+        // Vindicia screw up the paypal email field by a camel case typo in the response
+        // we need to deal with "payPalEmail" and "paypalEmail" in case they fix it on their side
```
nickyr: can you specify which is documented and which is what they actually return?

tests/Mock/FetchTransactionPayPalSuccess.xml
rondogency: yes

rondogency:
@nickyr fixed those small things you mentioned

src/ObjectHelper.php
```diff
@@ -50,6 +50,16 @@ public function buildTransaction(stdClass $object)
             }
         }
 
+        // Vindicia screw up the paypal email field by a camel case typo
```
parth21shah: Can you please reword this to something like
```
There is a discrepancy between Vindicia document and actual Vindicia response regarding 'paypalEmail`` field
```

nickyr: and can you say which is which?

PR #36, opened 2018-06-28
Add new fields in Payment Method object

rondogency:
Added type, postcode and country fields in Payment Method object, which is useful for getting paypal info of the transaction.

### DIFF
--- a/tests/PaymentMethodTest.php
+++ b/tests/PaymentMethodTest.php
@@ -105,6 +105,36 @@ class PaymentMethodTest extends TestCase
     /**
      * @return void
      */
+    public function testType()
+    {
+        $type = 'PayPal';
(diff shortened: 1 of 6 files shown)


tests/PaymentMethodTest.php
```diff
@@ -102,6 +102,36 @@ public function testCard()
...
     }
 
+    /**
+     * @return void
+     */
+    public function testType()
+    {
+        $type = 'PayPal';
+        $this->assertSame($this->paymentMethod, $this->paymentMethod->setType($type));
+        $this->assertSame($type, $this->paymentMethod->getType());
```
parth21shah: Can you also add a test in fetchPaymentMethodRequest with Type as a Payapal.

PR #39, opened 2018-08-09
Add reason code and case number field to Chargeback object

rondogency:
Add missing fields to be used in v18 upgrade.

### DIFF
--- a/src/Chargeback.php
+++ b/src/Chargeback.php
@@ -325,6 +325,10 @@ class Chargeback
 
     /**
      * Get the reason code
+     * 
+     * The reason code reported by your bank for this Chargeback object. 
+     * It's a 2-to-4-digit alphanumeric code provided by the issuing bank involved in a chargeback,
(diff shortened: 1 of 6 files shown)


src/Chargeback.php
```diff
@@ -322,4 +322,46 @@ public function setTransactionReference($value)
     {
         return $this->setParameter('transactionReference', $value);
     }
+
+    /**
+     * Get the reason code
```
nickyr: can you add an explanation and example of what this is?

src/Chargeback.php
```diff
@@ -322,4 +322,46 @@ public function setTransactionReference($value)
...
+     * @param string $value
+     * @return static
+     */
+    public function setReasonCode($value)
+    {
+        return $this->setParameter('reasonCode', $value);
+    }
+
+        /**
+     * Get the case number
```
nickyr: can you add an explanation and example of what this is?

PR #40, opened 2018-08-09
Added Omnipay support for Apple Pay

DanayaMel:
### What this PR does (product features, etc)

Added fields for Apple Pay payment. 

Vindicia | Type | Description
-- | -- | --
paymentData | Address | The paymentData field from PKPayment.
paymentInstrumentName | String | The paymentInstrumentName field from PKPayment
paymentNetwork | String | The paymentNetwork from PKPayment
transactionIdentifier | String | The transactionIdentifier field from PKPayment

### How it does that (implementation details)

Added getters and setters in NonStrippingCreditCard.php for payment fields. Also added tests for those fields in NonStrippingCreditCardTests.php.

**Closes:** #41579(the ticket(s) this PR fixes/closes)

**CI Build:** https://omnipay-support-applepay.ci.vimeows.com

## To Do
- N/A

### DB Alters
- N/A

### Standard pre-release checks
- [ ] Selenium checkout tests passed

### Things that could break
- N/A

### Things to QA
- [ ] Regression testing. Nothing should change.

### Crons and queues
_Fill in Y/N for each cron or queue in the "Need to disable?" column. During the release process, mark the remaining columns with X's as the steps are completed._

Cron/queue | Need to disable? (Y/N) | Disabled | Re-enabled
------------ | ------------- | ------------- | -------------
Transaction_SyncSales |N | |
VindiciaAutobill_StartAutobill |N | |
VindiciaAutobill_SaveAutobillsV6 |N | |
SyncProUser |N | |
SyncPlusUser |N | |
vimeo.cron.fetch_autobills.php | N| | 
vimeo.cron.fetch_transactions_new.php |N | |
vimeo.cron.plus_expiration.php |N | |
vimeo.cron.pro_expirations_new.php |N | |
vimeo.cron.live_expirations.php --acount_type='live_pro' |N | |
vimeo.cron.live_expirations.php --acount_type='live_business' | N| |
vimeo.cron.live_expirations.php --acount_type='live_premium' |N | |
vimeo.cron.live_expirations.php --acount_type='producer' | N| |

### People to contact
_Fill in Y/N for each group of people to denote who needs to be notified of this change. Fill in the last column with an X once they're ready._

Stakeholder | Need to notify? (Y/N) | Signed off
------------ | ------------- | -------------
BI | N| 
CRM | N| 
Community | N| 
Finance |N |

cc @Vimeo/transactions-team

### DIFF
--- a/src/NonStrippingCreditCard.php
+++ b/src/NonStrippingCreditCard.php
@@ -12,6 +12,13 @@ use Omnipay\Common\CreditCard;
 class NonStrippingCreditCard extends CreditCard
 {
     /**
+     * Internal storage of all of the card parameters.
+     *
+     * @var \Symfony\Component\HttpFoundation\ParameterBag
+     */
+    protected $parameters;
+
+    /**
      * Set Card Number
      *
      * Unlike the parent class, non-numeric characters are NOT stripped out of
@@ -27,5 +34,101 @@ class NonStrippingCreditCard extends CreditCard
          * @var static
          */
         return $this->setParameter('number', $value);
+    }
+
+    /**
+     * Sets the payment instrument name.
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setPaymentInstrumentName($value)
+    {
+        /**
+         * @var static
+         */
+        return $this->setParameter('paymentInstrumentName', $value);
+    }
+
+    /**
+     * Gets the payment instrument name.
+     *
+     * @return string|null
+     */
+    public function getPaymentInstrumentName()
+    {
+        return $this->getParameter('paymentInstrumentName');
+    }
+
+    /**
+     * Sets the payment network.
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setPaymentNetwork($value)
+    {
+        /**
+         * @var static
+         */
+        return $this->setParameter('paymentNetwork', $value);
+    }
+
+    /**
+     * Gets the payment network.
+     *
+     * @return string|null
+     */
+    public function getPaymentNetwork()
+    {
+        return $this->getParameter('paymentNetwork');
+    }
+
+    /**
+     * Sets the transaction identifier.
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setTransactionIdentifier($value)
+    {
+        /**
+         * @var static
+         */
+        return $this->setParameter('transactionIdentifier', $value);
+    }
+
+    /**
+     * Gets the transaction identifier.
+     * 
+     * @return string|null
+     */
+    public function getTransactionIdentifier()
+    {
+        return $this->getParameter('transactionIdentifier');
+    }
+
+    /**
+     * Sets the country, zip code, expiration date and account holder name.
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setPaymentData($value)
+    {
+        /**
+         * @var static
+         */
+        return $this->setParameter('paymentData', $value);
+    }
+
+    /**
+     * Gets the country, zip code, expiration date and account holder name.
+     *
+     * @return string|null
+     */
+    public function getPaymentData()
+    {
+        return $this->getParameter('paymentData');
     }
 }
--- a/tests/NonStrippingCreditCardTest.php
+++ b/tests/NonStrippingCreditCardTest.php
@@ -31,4 +31,52 @@ class NonStrippingCreditCardTest extends TestCase
         $this->assertSame($this->card, $this->card->setNumber($number));
         $this->assertSame($number, $this->card->getNumber());
     }
+
+    /**
+     * @return void
+     */
+    public function testPaymentInstrumentName()
+    {
+        $paymentInstrumentName = 'AmEx 1362';
+        $this->assertSame($this->card, $this->card->setPaymentInstrumentName($paymentInstrumentName));
+        $this->assertSame($paymentInstrumentName, $this->card->getPaymentInstrumentName());
+    }
+
+    /**
+     * @return void
+     */
+    public function testPaymentNetwork()
+    {
+        $paymentNetwork = 'AmEx';
+        $this->assertSame($this->card, $this->card->setPaymentNetwork($paymentNetwork));
+        $this->assertSame($paymentNetwork, $this->card->getPaymentNetwork());
+    }
+
+    /**
+     * @return void
+     */
+    public function testTransactionIdentifier()
+    {
+        $transactionIdentifier = '7e0c086a15c67fc8b573e4c43cde55c85785aadacd585ee66789a675fce116b1';
+        $this->assertSame($this->card, $this->card->setTransactionIdentifier($transactionIdentifier));
+        $this->assertSame($transactionIdentifier, $this->card->getTransactionIdentifier());
+    }
+
+    /**
+     * @return void
+     */
+    public function testPaymentData()
+    {
+        $paymentData = '{  
+         "version":"EC_v1",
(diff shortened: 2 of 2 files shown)


src/NonStrippingCreditCard.php
```diff
@@ -11,6 +11,13 @@
  */
 class NonStrippingCreditCard extends CreditCard
 {
+    /**
+     * Internal storage of all of the card parameters.
+     *
+     * @var \Symfony\Component\HttpFoundation\ParameterBag
+     */
+    protected $parameters;
```
nickyr: this is already defined on the parent so you don't need to redefine it here

src/NonStrippingCreditCard.php
```diff
@@ -28,4 +35,100 @@ public function setNumber($value)
...
+    public function setPaymentData($value)
+    {
+        /**
+         * @var static
+         */
+        return $this->setParameter('paymentData', $value);
+    }
+
+    /**
+     * Gets the country, zip code, expiration date and account holder name.
```
nickyr: where'd you find that this was what this field contains? the examples i've seen, including the one in the test below, look like a bunch of encrypted stuff.

do we actually need this field in Omnipay? if it is address information, we already have that elsewhere on the card object. if it's not address info, not sure it's super useful.

src/NonStrippingCreditCard.php
```diff
@@ -28,4 +35,100 @@ public function setNumber($value)
...
+        return $this->getParameter('paymentNetwork');
+    }
+
+    /**
+     * Sets the transaction identifier.
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setTransactionIdentifier($value)
```
nickyr: @parth21shah I thought we talked about not returning this since it didn't seem particularly useful and it's awkward to have a transaction-level ID on a payment method object? we've already got transaction IDs from vindicia

tests/NonStrippingCreditCardTest.php
```diff
@@ -31,4 +31,52 @@ public function testNumber()
         $this->assertSame($this->card, $this->card->setNumber($number));
         $this->assertSame($number, $this->card->getNumber());
     }
+
+    /**
+     * @return void
+     */
+    public function testPaymentInstrumentName()
+    {
+        $paymentInstrumentName = 'AmEx 1362';
```
nickyr: could you please use the faker to randomize the test data?

src/NonStrippingCreditCard.php
parth21shah: This is a token that ApplePay sends and I thought we need this to make sure when we are updating buildPayamentMethod it can handle all the parameters that we want to send or vindicia sends back

src/NonStrippingCreditCard.php
parth21shah: It is an address info but this field is only way we are going to get it

DanayaMel:
@nickyr @parth21shah  added new changes!

src/TestFramework/DataFaker.php
```diff
@@ -1076,4 +1076,52 @@ public function riskScore()
...
+     *
+     * @return array
+     */
+    public function paymentData()
+    {
+        return array(
+            'version' => $this->randomCharacters(self::ALPHABET_UPPER . self::DIGITS, $this->intBetween(1, 5)),
+            'data' => $this->randomCharacters(self::ALPHABET_UPPER . self::DIGITS, $this->intBetween(4, 120)),
+            'signature' => $this->randomCharacters(self::ALPHABET_UPPER . self::DIGITS, $this->intBetween(4, 20)),
+            'header' => array(
```
DanayaMel: Travis fails when I use the new array syntax.

nickyr: yeah omnipay supports php 5.2 which does not have new array syntax

src/NonStrippingCreditCard.php
nickyr: we only need to support parameters that are useful!

src/NonStrippingCreditCard.php
nickyr: if the address info really doesn't come back from vindicia other than in this field, we should parse this field in this library and set the information on the address fields on the card in the AbstractRequest::buildPaymentMethod function. so that would be part of the next PR, but this field still wouldn't be necessary on this card object.

DanayaMel:
@nickyr Changes made! Tests passed!

src/NonStrippingCreditCard.php
nickyr: i still don't think we need this parameter. is there any reason that we do?

src/NonStrippingCreditCard.php
```diff
@@ -28,4 +28,102 @@ public function setNumber($value)
...
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setToken($value)
+    {
+        /**
+         * @var static
+         */
+        return $this->setParameter('paymentData', $value);
```
nickyr: this should be `token` to match the function name

src/TestFramework/DataFaker.php
```diff
@@ -1076,4 +1076,52 @@ public function riskScore()
...
+     *
+     * @return string
+     */
+    public function paymentNetwork()
+    {
+        return $this->randomCharacters(self::ALPHABET_UPPER . self::ALPHABET_LOWER, $this->intBetween(4, 16));
+    }
+
+    /**
+     * Return payment data.
```
nickyr: can you make this more descriptive?

src/NonStrippingCreditCard.php
parth21shah: We have to send this in auth call otherwise it won't work. We decided not to save it on our side which is okay but have to tell this to vindicia

src/NonStrippingCreditCard.php
DanayaMel: I thought that parameter needed to match the object's name?

src/NonStrippingCreditCard.php
```diff
@@ -28,4 +28,102 @@ public function setNumber($value)
...
+     * Sets the payment instrument name.
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setPaymentInstrumentName($value)
+    {
+        /**
+         * @var static
+         */
```
parth21shah: Why do you specify this as static?

src/NonStrippingCreditCard.php
nickyr: @parth21shah the example you provided in the tech spec indicates that the transaction identifier is part of the payment data token already. do we really have to pass it to vindicia twice? do you have any docs about that?

DanayaMel: It doesn't look like the same transaction id.

DanayaMel: Changing the method names to setApplePayTransactionReference and getApplePayTransactionReference per discussion. May be able to eliminate it completely if need be, after further testing.

src/NonStrippingCreditCard.php
DanayaMel: No particular reason, I was following suit of setNumber.

DanayaMel: Which already existed in this class.

nickyr: it's because setParameter is defined in omnipay common, but is not typed correctly. so we have to override the typehint here to help psalm out

src/NonStrippingCreditCard.php
```diff
@@ -28,4 +28,104 @@ public function setNumber($value)
...
+     * @return string|null
+     */
+    public function getPaymentNetwork()
+    {
+        return $this->getParameter('paymentNetwork');
+    }
+
+    /**
+     * Sets Apple Pay's transaction reference.
+     * It's Apple Pay's reference to a transaction.
```
nickyr: this line may be redundant given the one above... probably more helpful to mention that apple calls this a transactionIdentifier

src/NonStrippingCreditCard.php
```diff
@@ -28,4 +28,104 @@ public function setNumber($value)
...
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setApplePayTransactionReference($value)
+    {
+        /**
+         * @var static
+         */
+        return $this->setParameter('transactionIdentifier', $value);
```
nickyr: should be applePayTransactionReference

src/NonStrippingCreditCard.php
nickyr: no, it actually needs to match the function name. the names of these parameters are only used by omnipay, they don't have anything to do with what comes back from vindicia, because we just take the stuff from vindicia and then pass it to this object, or pass it from this object to vindicia when building the request object.

when you're constructing a request for omnipay and pass it an array of string-indexed parameters, it actually uses the parameter name to identify which setters it should call for that object. so the parameter names actually have to match the setter names.

DanayaMel:
@nickyr @parth21shah Added changes!

src/TestFramework/DataFaker.php
```diff
@@ -1076,4 +1076,67 @@ public function riskScore()
...
+    }
+
+    /**
+     * Return a payment network.
+     *
+     * @return string
+     */
+    public function paymentNetwork()
+    {
+        return $this->randomCharacters(self::ALPHABET_UPPER . self::ALPHABET_LOWER, $this->intBetween(4, 16));
```
parth21shah: Does it make sense to randomly picking from array ['Visa', 'MasterCard', 'Amex']. So it's more inline with what we expect from Apple

PR #42, opened 2018-08-31
Add support for Apple Pay gateway

DanayaMel:
Added support for Apple Pay gateway.


src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,383 @@
...
+    public function getValidationUrl()
+    {
+        return $this->getParameter('validationURL');
+    }
+
+    /**
+     * Set the validation URL for the authorize request
+     *
+     * @param string $url
+     * @return static
```
DanayaMel: Not sure about what to do for this docblock. It's setting a parameter but psalms complaining about the return type being too general. I've tried using Omnipay\Message\AbstractRequest like it suggested but that didn't work either.

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,386 @@
...
+        // @TODO: Add merchantIdentifier and initiativeContext to private file and add here. 
+        $data["merchantIdentifier"] = "merchant.com";
+        $data["initiative"] = "web";
+        $data["initiativeContext"] = "abcd.com";
+
+        if (isset($this->data['validationURL'])) {
+            $data['validationURL'] = $this->getValidationUrl();
+        }
+
+        if (isset($this->data['epochTimestamp'])) {
```
nickyr: what is `$this->data` here? i don't think omnipay requests have a `data` attribute, so i think this is always going to return false

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,386 @@
...
+     * @return array
+     */
+    public function getData()
+    {
+        $this->validate('validationURL');
+
+        $data = array();
+
+        // These are from Apple's certificate – they should never change.
+        // @TODO: Add merchantIdentifier and initiativeContext to private file and add here.
```
nickyr: there can't be any private files in omnipay repos, private info would need to be passed as a parameter

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,386 @@
...
+        if (isset($this->data['expiresAt'])) {
+            $data['expiresAt'] = $this->getExpirationTimeStamp();
+        }
+
+        if (isset($this->data['merchantSessionIdentifier'])) {
+            $data['merchantSessionIdentifier'] = $this->getMerchantSessionID();
+        }
+
+        if (isset($this->data['nonce'])) {
+            $data['nonce'] = $this->getNonceToken();
```
nickyr: are all of these fields necessary to make the request?

src/Message/ApplePayAuthorizeRequest.php
nickyr: looks like that's what the other requests do... is it failing for all requests now due to a psalm update or something? if it's only failing on this one, that's a clue!

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,386 @@
...
+     *
+     * @return \Guzzle\Http\Message\RequestInterface
+     */
+    protected function createClientRequest($data, array $headers = null)
+    {
+        $config                          = $this->httpClient->getConfig();
+        $curlOptions                     = $config->get('curl.options');
+        $curlOptions[CURLOPT_SSLVERSION] = 6;
+        // @TODO: Add path to certs and passwords. 
+        $curlOptions[CURLOPT_SSLCERT] = "YOUR_CERTS_HERE";
```
nickyr: this needs to come from a parameter. perhaps it could be set when the gateway is constructed?

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: It's failing for all of them.

nickyr: OK guess psalm updated. The ideal thing would be for Omnipay\Common\Message\AbstractRequest::setParameter to have a return type of `static`, then we'd be good to go. But since it doesn't, we'll have to use a typehint to tell psalm what's going on. so basically you need to put a line above the return statement that says `/** @var static */`, which will tell psalm what the return type is actually supposed to be.

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,386 @@
...
+     */
+    protected static $RESPONSE_CLASS = '\Omnipay\Vindicia\Message\ApplePayAuthorizeResponse';
+
+    /**
+     * Default display name for Apple Pay Session.
+     * @TODO: Add displayname to private file and add here.
+     *
+     * @var string
+     */
+    const DEFAULT_DISPLAY_NAME = 'DISPLAY_NAME_HERE';
```
nickyr: this would also need to come from a parameter

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,386 @@
...
+        $httpRequest  = $this->createClientRequest($data);
+        $httpResponse = $httpRequest->send();
+
+        $statusCodes = array(
+            'message' => $httpResponse->getReasonPhrase(),
+            'statusCode' => $httpResponse->getStatusCode()
+        );
+
+        try {
+            $message = $httpResponse->json();
```
nickyr: variable naming here is a bit confusing because `$statusCodes['message']` is `$httpResponse->getReasonPhrase()` but `$message` is  `$httpResponse->json()`. So it's not clear what the difference between the two messages is

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,386 @@
...
+            $response = array_merge(
+                $statusCodes,
+                $message
+            );
+        // If you try to parse an empty response body, error will be thrown.
+        } catch (\RunTimeException $e) {
```
nickyr: is it possible to detect the empty response body before calling `json()` rather than doing exception handling?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,386 @@
+<?php
+
+namespace Omnipay\Vindicia\Message;
+
+use Omnipay\Common\Message\ResponseInterface;
+
+class ApplePayAuthorizeRequest extends \Omnipay\Common\Message\AbstractRequest
```
nickyr: can you add a classblock explaining how to use this? see the other requests for examples

nickyr:
Alright, finished going through now. Looks generally good, and the request sending seems good!

The two main things to address are:

1. Certificates/passwords/etc. Whoever is using this library needs to be able to pass this stuff in as a parameter--they won't be able to edit the library to stick stuff in. I think probably the way to go is to have the certificate passed in when the gateway is instantiated, along with the vindicia username and password.
2. Data getting passed to the request. It seems like there are more getters and setters than how many variables I would expect we actually need to set on the request. For example, I'm confused as to why we would need to set epochTimestamp. Could you go through those variables and make sure you're just providing getters and setters for the ones that are useful/necessary? That'll make it a lot easier to use this gateway.

DanayaMel:
@nickyr  I thought we could use the timestamps to detect when a payment session would end. Since the user only has a specified amount of time to make a purchase.

nickyr:
@DanayaMel are we allowed to specify the amount of time a user has?

src/Message/ApplePayAuthorizeRequest.php
nickyr: this is on public github, so we can't specify vimeo anything anywhere

DanayaMel:
> @DanayaMel are we allowed to specify the amount of time a user has?

Nope.

nickyr:
@DanayaMel in that case, we don't need getters and setters on the request object for that time, because that implies that we can change it when we actually can't

we only need to be able to set things on the request if they'll actually do something. does that make sense?

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,164 @@
...
+    {
+        $transaction = $this->getTransaction();
+        if (isset($transaction)) {
+            return $transaction->getPayPalRedirectUrl();
+        }
+        return null;
+    }
+
+    /**
+     * @return string|null
```
nickyr: could you add docs explaining what all of these things you can get mean? also, if they're not actually useful, you could just delete them

DanayaMel: Going to add this in the next coming commit.

src/ApplePayGateway.php
```diff
@@ -0,0 +1,159 @@
+<?php
+
+namespace Omnipay\Vindicia;
+
+
+class ApplePayGateway extends AbstractVindiciaGateway
```
nickyr: could you add a classblock here explaining how to use this gateways, like the other gateways have?

src/ApplePayGateway.php
```diff
@@ -0,0 +1,159 @@
...
+{
+
+    /**
+     * Get the gateway parameters.
+     *
+     * @return array
+     */
+    public function getDefaultParameters()
```
nickyr: this is the same as the base class, so probably not necessary?

nickyr: or maybe the cert parameters should be added here

src/ApplePayGateway.php
```diff
@@ -0,0 +1,159 @@
...
+     */
+    public function getPassword()
+    {
+        return $this->getParameter('password');
+    }
+
+    /**
+     * @return AbstractVindiciaGateway
+     */
+    public function setPassword($value)
```
nickyr: these are all in the base class

src/ApplePayGateway.php
```diff
@@ -0,0 +1,159 @@
...
+    public function getPemCertPath()
+    {
+        return $this->getParameter('pemCertPath');
+    }
+
+    /**
+     * @param string $value
+     * @return static
+     */
+    public function setPemCertPath($value)
```
nickyr: could you add docs explaining what these certs are and what the path should look like?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+
+/**
+ * Retrieve an Apple Pay Payment Session object.
+ *
+ * Post a request to Apple Pay server's Payment Session endpoint using two-way TLS.
+ *
+ * Parameters:
+ * - pemCertPath: Path to the cert of the split Merchant Identity certification,
+ * associated with your merchantID needed to make the call.
+ * - keyCertPath: Path to the key cert of the split Merchant Identity certification,
```
nickyr: these are gateway parameters, not request parameters, right? they don't need to be documented with the request if so

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+ * associated with your merchantID needed to make the call.
+ * - keyCertPassword: Key certification password for the key certification.
+ * - validationUrl: Validation URL received from session.onvalidatemerchant on the client.
+ * - merchantIdentifier: Your merchant ID provided by Apple.
```
nickyr: is this the same for every request? maybe it should be on the gateway too. or could you conceivably have multiple?

and how about for applicationUrl below?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+ * - displayName: A name for your store, suitable for display, appears in the touch bar.
+ * - applicationType: The "initiative" parameter depending on the type of application.
+ * For Apple Pay on the web, use "web". For Business Chat, use "messaging".
+ * - applicationUrl: The initive context is based on the value supplied for the applicationType/initiative.
```
nickyr: initiative is spelled wrong i believe

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+ * For Apple Pay on the web, provide your fully qualified domain name associated with your
```
nickyr: could you provide an example of this?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+ * <code>
+ *   // set up the gateway
+ *   $gateway = \Omnipay\Omnipay::create('Vindicia ApplePay');
+ *   $gateway->setUsername('your_username');
+ *   $gateway->setPassword('y0ur_p4ssw0rd');
+ *   $gateway->setPemCertPath('./certs/path_to_cert.crt.pem');
+ *   $gateway->setKeyCertPath('./certs/path_to_key_cert.key.pem');
+ *   $gateway->setKeyCerPassword('y0ur_key_p4ssw0rd');
+ *   $gateway->setTestMode(false);
+ *
```
nickyr: can you add a comment about the front end step that has to happen before you can call authorize?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+ *   // authorize the request and get an Apple Pay Session object.
+ *   $ApplePaySessionObject = $gateway->authorize(array(
+ *       "validationUrl" => "https://example.com",
```
nickyr: this is an apple URL that has a very specific structure, right? could you provide a more realistic example?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+    {
+        $pemCerts = $this->getPemCertPath();
+        $keyCerts = $this->getKeyCertPath();
+        $keyCertPassword = $this->getKeyCertPassword();
+
+        // Setting options for the request.
+        $config                          = $this->httpClient->getConfig();
+        $curlOptions                     = $config->get('curl.options');
+        $curlOptions[CURLOPT_SSLVERSION] = 6;
+        $curlOptions[CURLOPT_SSLCERT] = $pemCerts;
```
nickyr: could you keep the spacing style consistent? either lining up the equals signs or just having one space is fine, but a mix is confusing

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,164 @@
...
+     * Is the response successful?
+     * Throws an exception if there's no code.
+     *
+     * @return boolean
+     * @throws \Omnipay\Common\Exception\InvalidResponseException
+     */
+    public function isSuccessful()
+    {
+        $statusCode = $this->getStatusCode();
+        return ($statusCode >= 200 && $statusCode < 300) || $statusCode == 304;
```
nickyr: what does a 304 mean? and can they really return that many different 200s?

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Is the response successful?
+     * Throws an exception if there's no code.
```
nickyr: what do you mean if there's no code? in general, having these functions return null instead of throwing an exception is what omnipay generally does

nickyr: i don't see where the exception is being thrown tho

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getRedirectUrl()
+    {
+        $transaction = $this->getTransaction();
+        if (isset($transaction)) {
+            return $transaction->getPayPalRedirectUrl();
```
nickyr: paypal?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+ *   $gateway = \Omnipay\Omnipay::create('Vindicia ApplePay');
+ *   $gateway->setUsername('your_username');
+ *   $gateway->setPassword('y0ur_p4ssw0rd');
+ *   $gateway->setPemCertPath('./certs/path_to_cert.crt.pem');
+ *   $gateway->setKeyCertPath('./certs/path_to_key_cert.key.pem');
+ *   $gateway->setKeyCerPassword('y0ur_key_p4ssw0rd');
+ *   $gateway->setTestMode(false);
+ *
+ *   // authorize the request and get an Apple Pay Session object.
+ *   $ApplePaySessionObject = $gateway->authorize(array(
```
nickyr: it returns a response, right? not an apple pay object?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,362 @@
...
+ *       "validationUrl" => "https://example.com",
+ *       "merchantIdentifier" => "merchant.com.example",
+ *       "displayName" => "MyStore",
+ *       "applicationType" => "type", // initiative: "web" or "messaging"
+ *       "applicationUrl" => "mystore.example.com" // initiativeContext:  domain name or payment gateway URL
+ *       )
+ *   ))->send();
+ *
+ * Pass this object back to the client to validate your merchant and continue with an Apple Pay payment.
```
nickyr: i suspect you don't mean to pass an Omnipay response object back from the client... can you specify exactly what needs to be passed back and provide an example of retrieving that from the response?

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,164 @@
...
+            return $this->data['merchantSessionIdentifier'];
+        }
+
+        return null;
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getApplePaySessionNonceToken()
```
nickyr: can you add docs explaining what all these apple pay specific things are?

src/ApplePayGateway.php
DanayaMel: I'm going to get rid of it because I don't believe it's necessary. Also, adding the certs here throws an error because getPemCertPath(), etc. would need to be in the capture and completeAuthorize methods. We don't need them for that.

src/ApplePayGateway.php
DanayaMel: Right, forgot to remove these. Was trying to see if I could add them here so that I wouldn't have to in authorize(), but it checks for them in that method so... eh

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: They are gateway parameters but I thought they should be documented here because we're using them in the request object. Also, it would be less confusing as its in the example object that Apple Requests in the API docs: https://developer.apple.com/documentation/apple_pay_on_the_web/apple_pay_js_api/requesting_an_apple_pay_payment_session

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: I think you could have multiple, depends on the context. I'd rather leave them here since they're so specific to the context of the authorize request being sent.

src/Message/ApplePayAuthorizeResponse.php
DanayaMel: Copy and paste mistake. Fixing this.

src/Message/ApplePayAuthorizeResponse.php
DanayaMel: Whoops, copy and paste monster got me.

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: I will need to connect all the parts to verify how I would send it back. Essentially, what you get back is what you need for the Apple Pay Session.

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: See comment above

src/Message/ApplePayAuthorizeResponse.php
DanayaMel: Sure, they are all what you get back from Apple. I'll add a comment.

src/ApplePayGateway.php
DanayaMel: It's a path to the user's personal path, so I'm not sure how I would show what that would look like... I've used `('./certs/path_to_cert.crt.pem');` in the main docblock. Also, there really isn't much information on what the certs are besides the fact that you get them from apple and have to split them. Not sure if I should add that since it's not apart of Appple Pay docs. I got my information from scouring forums/stackoverflow.

src/Message/ApplePayAuthorizeRequest.php
nickyr: OK, can you specify that they get set by the gateway then?

src/Message/ApplePayAuthorizeRequest.php
nickyr: ok sounds good

src/Message/ApplePayAuthorizeRequest.php
nickyr: right, but the authorize function returns an Omnipay response. So there's something specific in the Omnipay response that actually gets passed to the front end. You'll need to figure out what that is in this PR so that you can make sure it is accessible on the response and named and documented appropriately.

src/ApplePayGateway.php
nickyr: ^^ that example is helpful, you could add it here too

src/Message/ApplePayAuthorizeResponse.php
DanayaMel: I'll just leave it at 200 for now. If any other codes pop up we can update it.

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: Well essentially the front end should get the entire data object. I'm thinking I could have a function to return all the of the data.

nickyr: Yeah if the front end needs an entire object that's returned by apple, than having a function to retrieve that object from the omnipay response makes sense

src/Message/ApplePayAuthorizeResponse.php
DanayaMel: Since I'm returning the Apple Pay Payment Session object as one json object, I no longer think these individual pieces of it are necessary. Apple knows how to handle expired session objects anyway.

nickyr: perfect!

DanayaMel:
@nickyr added changes!

src/ApplePayGateway.php
```diff
@@ -0,0 +1,162 @@
...
+ * <code>
+ *   // set up the gateway
+ *   $gateway = \Omnipay\Omnipay::create('Vindicia_ApplePay');
+ *   $gateway->setUsername('your_username');
+ *   $gateway->setPassword('y0ur_p4ssw0rd');
+ *   $gateway->setPemCertPath('./certs/path_to_cert.crt.pem');
+ *   $gateway->setKeyCertPath('./certs/path_to_cert.key.pem');
+ *   $gateway->setKeyCertPassword('your_key_cert_password');
+ *   $gateway->setTestMode(false);
+ *
```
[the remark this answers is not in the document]
DanayaMel: That's why I wanted to show the flow with the standard ApplePayJS docs code.

With some of your suggestions, I'm worried about these docs becoming ApplePayJS tutorials, because that's not what this is for. It's just one piece of the whole API and can be implemented many different ways. I like the suggestion of listing what should happen "user clicks button, payment sheet is loaded... etc" and then where parts of the gateway fits in.

I also can't go into detail about the completeAuthorize and capture because I haven't done that yet. This is one of those issues we spoke about with the tickets, I'm going to have to come back to this one later. So, I can make most of the changes you've suggested but I'm going to have to come back to this and update it once I've gone through the entire flow.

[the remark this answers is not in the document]
nickyr: I agree that this should not be an ApplePayJS tutorial. If someone were following an ApplePayJS tutorial, they just need to be able to tell roughly where these different steps are supposed to fit in the code. So yeah, just adding brief notes about what is supposed to happen on the front end side is totally fine.

Sure, that's fine to add the docs for completeAuthorize and capture later

src/ApplePayGateway.php
```diff
@@ -0,0 +1,162 @@
...
+ *   // The Apple Pay Payment Session session (expires after 5 mins)
+ *   if ($authorizeResponse->isSuccessful()) {
+ *       echo "Status Reason: " . $authorizeResponse->getReason() . PHP_EOL;
+ *       echo "Status Code: " . $authorizeResponse->getStatusCode() . PHP_EOL;
+ *       echo "Expiration time: " . $authorizeResponse->getApplePaySessionExpirationTimeStamp() . PHP_EOL;
+ *       echo "Apple Pay Payment Session Object: " . $authorizeResponse->getApplePayPaymentSessionObject(). PHP_EOL;
+ *   } else {
+ *       // error handling
+ *   }
+ *
```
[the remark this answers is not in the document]
DanayaMel: Regarding the capture and completeAuthorize methods, can I add todo comments for now as place holders? I can remove them when I have to update these functions later.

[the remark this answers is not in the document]
nickyr: sure

DanayaMel:
@nickyr Added changes!

tests/Message/ApplePayAuthorizeRequestTest.php
```diff
@@ -0,0 +1,147 @@
...
+	*/
+    public function testSendSuccess()
+    {
+        $this->setMockHttpResponse('ApplePayAuthorizeRequestSuccess.txt');
+        $response = $this->request->send();
+
+        $this->assertTrue($response->isSuccessful());
+        $this->assertSame('OK', $response->getReason());
+        $this->assertSame(200, $response->getStatusCode());
+    }
```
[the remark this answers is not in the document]
DanayaMel: Adding this change now.

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,72 @@
...
+     * This should be passed to the front end.
+     *
+     * @return static
+     */
+    public function getPaymentSessionObject()
+    {
+        /**
+         * @var static
+         */
+        return (empty($this->data['body'])) ? '' : json_encode($this->data['body']);
```
DanayaMel: I have to do this because json encoding the empty body becomes "''".

DanayaMel: @nickyr

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,462 @@
...
+ * Parameters set by this request:
+ * - validationUrl: Validation URL received from session.onvalidatemerchant on the client.
+ * - merchantIdentifier: Your merchant ID provided by Apple.
+ * - displayName: A name for your store, suitable for display, appears in the touch bar.
+ * - applicationType: The "initiative" parameter depending on the type of application:
+ *   For Apple Pay on the web, use "web".
+ * - applicationUrl: The initiative context is based on the value supplied for the applicationType/initiative:
+ *   Provide your fully qualified domain name associated with your Apple Pay Merchant Identity Certificate.
+ *
+ * See @link for more details.
```
nickyr: what's this referencing?

src/Message/ApplePayAuthorizeRequest.php
nickyr: could you update this to match the docs in the gateway file? the gateway file uses abstractResponse as the name here, which is more correct, since it returns a response, not an apple pay object

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,421 @@
...
+    /**
+     * @return null|string
+     */
+    public function getKeyCertPassword()
+    {
+        return $this->getParameter('keyCertPassword');
+    }
+
+    /**
+     * Password associated with your key certificate, if applicable.
```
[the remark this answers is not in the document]
nickyr: can you please address this comment?^^

src/ApplePayGateway.php
nickyr: i like the docs you added for these files in the request, could you add them here too?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,462 @@
...
+            );
+        }
+
+        /**
+         * @var ApplePayAuthorizeResponse
+         */
+        $this->response = new static::$RESPONSE_CLASS($this, $response);
+
+        /**
+         * @var ApplePayAuthorizeResponse
```
nickyr: only one of these @var annotations is necessary since they're annotating the same variable to be the same type

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,72 @@
...
+         * @var boolean
+         */
+        return ($this->data['statusCode'] == 200);
+    }
+
+    /**
+     * The entire Apple Pay payment session object to be used to validate the merchant.
+     * This should be passed to the front end.
+     *
+     * @return static
```
nickyr: i don't think this return type is correct. static would be a Response object, but it looks like it returns a string

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,72 @@
...
+        /**
+         * @var static
+         */
+        return (empty($this->data['body'])) ? '' : json_encode($this->data['body']);
+    }
+
+    /**
+     * The human-readable status code.
+     *
+     * @return static
```
nickyr: same here, this is a string, not a Response

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public function getReason()
```
nickyr: Omnipay calls this "message", not "reason". see https://github.com/vimeo/omnipay-vindicia/blob/master/src/Message/Response.php

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,72 @@
...
+        /**
+         * @var static
+         */
+        return $this->data['reasonPhrase'];
+    }
+
+    /**
+     * The numerical status code.
+     *
+     * @return static
```
nickyr: this is a string, not a Response

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public function getStatusCode()
```
nickyr: Omnipay calls this "code", not "statusCode". see https://github.com/vimeo/omnipay-vindicia/blob/master/src/Message/Response.php

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,72 @@
...
+         */
+        return $this->data['statusCode'];
+    }
+
+    /**
+     * The entire response with status reason and phrase.
+     *
+     * @return static
+     */
+    public function getResponse()
```
nickyr: what is this going to be used for? returning a string of the entire response kinda defeats the purpose of omnipay because then it would be different for every gateway.

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,462 @@
...
+        return $this->getParameter('applicationType');
+    }
+
+    /**
+     * Sets the application type or the "initiative" for the request
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setApplicationType($value)
```
nickyr: since you deleted all the business chat stuff, having the application type parameter isn't necessary because it will always just be "web". so i think you could just delete all mention of this parameter and pass "web" directly

tests/Message/ApplePayAuthorizeRequestTest.php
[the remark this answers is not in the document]
nickyr: doesn't look like this ever got added?

tests/Message/ApplePayAuthorizeRequestTest.php
```diff
@@ -0,0 +1,150 @@
...
+
+   /**
+	* @return void
```
nickyr: funky indentation

tests/Message/ApplePayAuthorizeResponseTest.php
```diff
@@ -0,0 +1,134 @@
...
+namespace Omnipay\Vindicia\Message;
+
+use Omnipay\Vindicia\TestFramework\Mocker;
+use Omnipay\Vindicia\TestFramework\DataFaker;
+use Omnipay\Tests\TestCase;
+
+class ApplePayAuthorizeResponseTest extends TestCase
+{
+   /**
+	* @return void
```
nickyr: funky indentation

tests/Message/ApplePayAuthorizeRequestTest.php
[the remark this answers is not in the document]
nickyr: oh i see you added this in ApplePayAuthorizeResponseTest.php. Omnipay typically just tests methods on the response in the RequestTests, because they have a lot of the same boilerplate and what's important is really that you can get the appropriate data after making the request. so i'd recommend just combining the two files.

src/ApplePayGateway.php
DanayaMel: Did you mean like this part of the request doc?

https://github.com/vimeo/omnipay-vindicia/pull/42/files#diff-ba17822789aa8e485b85ea6429963389R16

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: This link tag in the code:

https://github.com/vimeo/omnipay-vindicia/pull/42/files#diff-ba17822789aa8e485b85ea6429963389R88

src/ApplePayGateway.php
nickyr: I was thinking https://github.com/vimeo/omnipay-vindicia/pull/42/files#diff-ba17822789aa8e485b85ea6429963389R148

src/Message/ApplePayAuthorizeRequest.php
nickyr: oh maybe just move those four lines up here? might be more clear

src/ApplePayGateway.php
```diff
@@ -0,0 +1,207 @@
...
+ * Vindicia ApplePay Gateway
+ * This Apple Pay Gateway provides the functionality to authorize, complete authorize and capture
+ * Apple Pay payments for Apple Pay on the web.
+ * 
+ * // TODO: Add documentation about functionality for capture and completeAuthorize methods to capture
+ * // and authorize payments.
+ * 
+ * Apple Pay is available on all iOS devices with a Secure Element—an industry-standard, certified
+ * chip designed to store payment information safely. On macOS, users must have anApple Pay-capable
+ * iPhone or Apple Watch to authorize the payment, or a MacBook Pro
```
nickyr: line break is a bit too early

src/ApplePayGateway.php
```diff
@@ -0,0 +1,207 @@
...
+ * with Touch ID.
+ *
+ * You use the same username and password as you do with the regular Vindicia gateway.
+ *
+ * Optional parameters set by this gateway:
```
nickyr: are these really optional? i thought you needed the cert

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,443 @@
...
+
+/**
+ * Retrieve an Apple Pay Payment Session object.
+ *
+ * Post a request to Apple Pay server's Payment Session endpoint using two-way TLS.
+ * The response to this request is an Apple Pay session object – it expires after five minutes.
+ * Pass the session object to the completion function completeMerchantValidation in the front end
+ * to verify the merchant.
+ *
+ * Optional parameters set by the gateway and this class:
```
nickyr: as above, are they actually optional?

src/Message/ApplePayAuthorizeRequest.php
nickyr: this still needs to be addressed. i'd recommend just copying the example from the gateway to here because i think that one is all up to date

src/Message/ApplePayAuthorizeResponse.php
```diff
@@ -0,0 +1,59 @@
...
+     * Was the response successful?
+     *
+     * @return boolean
+     */
+    public function isSuccessful()
+    {
+        /**
+         * @var boolean
+         */
+        return ($this->data['code'] == '200');
```
nickyr: could use $this->getCode() here

src/TestFramework/DataFaker.php
```diff
@@ -1038,7 +1038,7 @@ public function soapId()
         } while ($result == 0);
         return $result;
     }
-
+
```
nickyr: no need for all these spaces!

tests/Message/ApplePayAuthorizeRequestTest.php
```diff
@@ -0,0 +1,176 @@
...
+
+        $this->assertFalse($response->isSuccessful());
+        $this->assertSame('400', $response->getCode());
+        $this->assertSame('Not Found', $response->getMessage());
+    }
+
+    /**
+     * @return void
+     */
+    public function testGetPaymentSessionObjectFailure()
```
nickyr: these two tests could just be combined with the two tests above. this one could be part of testSendFailure and testGetPaymentSessionObjectSuccess could be part of testSendSuccess. That's how most omnipay repos do it

src/ApplePayGateway.php
DanayaMel: Well the cert password certainly is.

Technically, I split the cert into a key and a pem but I believe there may be a way to only use one cert. So, I put that they were optional since you can send the request with or without a cert.

nickyr: if you send the request without a cert, what happens? does apple accept that? how do they know it's real?

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: Oh, I guess I didn't fully understand what you meant. It's the gateway code example that you want in the request docs? I thought to keep them separate since it's in the request.

src/ApplePayGateway.php
DanayaMel: Apple sends an error. I can reword it, I wanted the user to know that they were optional incase they had a different cert setup.

src/Message/ApplePayAuthorizeRequest.php
nickyr: Well what I mean is just that you're still calling this $ApplePaySessionObject, not $authorizeResponse. Although you're calling it $authorizeResponse below. And you've got it right in the gateway request docs. So just fixing that is fine.

src/ApplePayGateway.php
nickyr: This library doesn't support a different cert setup though, right? It may be optional for Apple Pay, but not if you're using this library?

DanayaMel: There's definitely more than one cert setup for Apple Pay though. The way I did it to get it to work is not listed on Apple's website.

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: Ohhh, okay, gotcha now.

src/ApplePayGateway.php
nickyr: If you were a user of this library, how would you make it work with a different cert setup? You're writing documentation for this omnipay library, not apple pay, and it looks to me like this library needs the two certs in order to successfully make an apple pay transaction.

If I'm missing something and there is a different set of parameters that you could specify to make a successful apple pay transaction, can you let me know what those are?

DanayaMel: It's really just pem cert and an additional key cert vs a pem cert that contains a key cert in it as well.

DanayaMel: I never got it to work, that is the way apple intended it though.

nickyr: sounds good! the docs just need to say one way to use it, it's fine if we don't provide every possible way to use it

DanayaMel:
@nickyr All changes have been made!

src/ApplePayGateway.php
```diff
@@ -0,0 +1,206 @@
...
+    {
+        return $this->setParameter('keyCertPath', $value);
+    }
+
+    /**
+     * @return null|string
+     */
+    public function getKeyCertPassword()
+    {
+        return $this->getParameter('keyCertPassword');
```
JCManzo: Could you make all of these strings into consts (keyCertPassword, keyCertPassword, pemCertPath)

src/ApplePayGateway.php
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Password associated with your key certificate.
+     * Can be omitted if a password isn't set.
+     * 
+     * @param string $value
+     * @return static
+     */
+    public function setKeyCertPassword($value)
```
JCManzo: Could you add typehints to all function params and return values. Some functions have them and others don't so better to be consistent.

src/ApplePayGateway.php
```diff
@@ -0,0 +1,206 @@
...
+        return $this->setParameter('keyCertPassword', $value);
+    }
+
+    /**
+     * Makes request to Apple to set up session between Vimeo and Apple.
+     *
+     * @param array $parameters
+     * @return \Omnipay\Vindicia\Message\ApplePayAuthorizeRequest
+     */
+    public function authorize(array $parameters = array())
```
JCManzo: What are all the possible valid parameter values here?  I know you have an example in the class docblock but that would require the user of this function to scroll up and go looking for it. I think it would be better to keep related information close to where it will be used.

If you make all these param values in `ApplePayAuthorizeRequest.php` into consts, you could reference them from here. Just an idea.

src/ApplePayGateway.php
```diff
@@ -0,0 +1,206 @@
...
+{
+
+    /**
+     * Get the gateway name.
+     *
+     * @return string
+     */
+    public function getName()
+    {
+        return 'Vindicia ApplePay';
```
JCManzo: This could be a constant as well?

src/ApplePayGateway.php
```diff
@@ -0,0 +1,206 @@
...
+        return $this->createRequest('\Omnipay\Vindicia\Message\ApplePayAuthorizeRequest', $parameters);
+    }
+
+    /**
+     * Authorize an Apple Pay purchase.
+     *
+     * @param array $parameters
+     * @return \Omnipay\Common\Message\AbstractRequest
+     */
+    public function completeAuthorize(array $parameters = array())
```
JCManzo: Same issue here about all possible valid $parameter values.

src/ApplePayGateway.php
```diff
@@ -0,0 +1,206 @@
...
+        return $this->createRequest('\Omnipay\Vindicia\Message\AuthorizeRequest', $parameters);
+    }
+
+    /**
+     * Capture an Apple Pay purchase.
+     *
+     * @param array $parameters
+     * @return \Omnipay\Vindicia\Message\CaptureRequest
+     */
+    public function capture(array $parameters = array())
```
JCManzo: and here.

src/ApplePayGateway.php
```diff
@@ -0,0 +1,206 @@
...
+ *    // getApplePayPaymentSessionObject() can then be called to retrieve the ApplePayPaymentSession object.
+ *    // You can ensure the response is successful before sending it to the front end. If there is an error the
+ *    // response body will be empty.
+ *    if ($authorizeResponse->isSuccessful()) {
+ *        echo 'Status Code: ' . $authorizeResponse->getMessage() . PHP_EOL;
+ *        echo 'Status Reason: ' . $authorizeResponse->getMessage() . PHP_EOL;
+ *        echo 'Apple Pay Payment Session Object: ' . $authorizeResponse->getPaymentSessionObject()
+ *          . PHP_EOL;
+ *        echo 'Response: ' . $authorizeResponse->getResponse() . PHP_EOL;
+ *    } else {
```
JCManzo: Else needs to be in next line.

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,443 @@
...
+     * @var string
+     */
+    protected static $RESPONSE_CLASS = '\Omnipay\Vindicia\Message\ApplePayAuthorizeResponse';
+
+    /**
+     * @return null|string
+     */
+    public function getUsername()
+    {
+        return $this->getParameter('username');
```
JCManzo: Could you make all these parameter strings in this class into constants as well?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,443 @@
...
+        return 'POST';
+    }
+
+    /**
+     * @return array
+     */
+    public function getData()
+    {
+        // We must have the validation URL set. If it isn't, don't make the call.
+        $this->validate('validationUrl');
```
JCManzo: Should we throw an exception here if the validation URL is not set rather than relying on the user to have set the value before calling `getData()`?

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,443 @@
...
+        return $data;
+    }
+
+    /**
+     * @param       $data
+     * @param array $headers
+     *
+     * @return static
+     */
+    protected function createClientRequest($data, array $headers = null)
```
JCManzo: Please add type hints to all function params and return types to be consistent. Some docbloc params are also missing their types.

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,443 @@
...
+        $curlOptions[CURLOPT_SSLVERSION]   = 6;
+        $curlOptions[CURLOPT_SSLCERT]      = $pemCertPath;
+        $curlOptions[CURLOPT_SSLKEY]       = $keyCertPath;
+        $curlOptions[CURLOPT_SSLKEYPASSWD] = $keyCertPassword;
+        $curlOptions[CURLOPT_POST]         = 1;
+
+        $config->set('curl.options', $curlOptions);
+        $this->httpClient->setConfig($config);
+
+        // don't throw exceptions for 4xx errors
```
JCManzo: Why are be suppressing errors?

src/TestFramework/DataFaker.php
```diff
@@ -1145,6 +1145,20 @@ public function applePayTransactionReference()
...
+     *
+     * @return string
+     */
+    public function path()
+    {
+        return '/'
+        . self::ALPHABET_LOWER
+        . '/'
+        . self::ALPHABET_LOWER
+        . '/';
```
JCManzo: these string appends should be indented 4 more spaces each

src/TestFramework/DataFaker.php
```diff
@@ -1145,6 +1145,20 @@ public function applePayTransactionReference()
         return $result;
     }
 
+    /**
+     * Return a path.
```
JCManzo: What kind of path? What is this path for?

tests/ApplePayGatewayTest.php
```diff
@@ -0,0 +1,127 @@
...
+        $this->displayName = $this->faker->username();
+        $this->applicationUrl = $this->faker->url();
+    }
+
+    /**
+     * @return void
+     */
+    public function testGetName()
+    {
+        $this->assertSame('Vindicia ApplePay', $this->gateway->getName());
```
JCManzo: Once you make this string into a constant in the corresponding class, you can update this to use that.

src/ApplePayGateway.php
DanayaMel: I don't think that would be necessary since there's a lot of them, mostly only used twice. Also, since this is an extension of the Omnipay-Common library, I believe we're trying to keep the same style format as the original library as well as the other classes in this repo.

Omnipay-Common found here: https://github.com/thephpleague/omnipay-common

src/ApplePayGateway.php
DanayaMel: I'm not sure that will work because the value of these setters depends on the context of the person using them.

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: This is because Guzzle throws an error for a 400 status code. When that happens, we aren't able to get the ApplePayAuthorizeResponse we want, which should be returned no matter what the status code is.

src/TestFramework/DataFaker.php
DanayaMel: This is to fake a path to a file. For my case I need this to test getting and setting pem/key certs. I'm making it general so that it can be used for a different context in the future.

I used it here: https://github.com/vimeo/omnipay-vindicia/pull/42/files#diff-fa49bbd6059ef71c54689eb53f1b3f1eR18

DanayaMel: I can update the doc block to make it more clear.

src/ApplePayGateway.php
nickyr: Omnipay style is to do it like this

src/ApplePayGateway.php
nickyr: the normal omnipay documentation style is to document them in the request file (ApplePayAuthorizeRequest.php). it could be useful to add a note here to direct people to that file for the parameters. that'd be helpful to people less familiar with omnipay.

src/ApplePayGateway.php
nickyr: do you mean php 7 style typehints? or @param annotations? omnipay supports php 5, so we can't use the php 7 style typehints. (`array` is supported in php 5.) if there are any missing @param annotations, those should definitely be addressed. just let danaya know where you saw them

src/ApplePayGateway.php
DanayaMel: Don't think this is necessary since it's only instance is here.

src/ApplePayGateway.php
DanayaMel: I added a TODO in the docblock to add documentation and functionality for completeAuthorize() and capture() later.

src/Message/ApplePayAuthorizeRequest.php
DanayaMel: That's sort of what this validate function does. It will try to get the passed in data for the request, if that parameter isn't set, it will throw an error.

DanayaMel: I will update the comment, don't think that was clear either.

tests/ApplePayGatewayTest.php
DanayaMel: See comment above ^^. For this test, I believe it should be hardcoded, to make sure I'm getting exactly what I'm expecting. If there was a typo in the gateway class that this tests, this would catch that.

DanayaMel:
@JCManzo all changes made!

src/ApplePayGateway.php
nickyr: works for me, this is how Omnipay typically does it

src/ApplePayGateway.php
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Vindicia ApplePay Gateway
+ * This Apple Pay Gateway provides the functionality to authorize, complete authorize and capture
+ * Apple Pay payments for Apple Pay on the web.
+ * 
+ * // TODO: Add documentation about functionality for capture and completeAuthorize methods to capture
+ * // and authorize payments.
+ * 
+ * Apple Pay is available on all iOS devices with a Secure Element—an industry-standard, certified
```
rondogency: please fix the spacing in docs

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,443 @@
...
+        return $data;
+    }
+
+    /**
+     * @param       $data
+     * @param array $headers
+     *
+     * @return static
+     */
+    protected function createClientRequest($data, array $headers = null)
```
rondogency: Please add docblock for this function, and it would be better if you could write down what kind of role this is playing in the payment flow

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -0,0 +1,443 @@
...
+        /**
+         * @var static
+         */
+        return $httpRequest;
+    }
+
+    /**
+     * Overriding AbstractRequest::sendData() so that we can make a REST call instead of a SOAP call.
+     * @param array $data
+     * @psalm-suppress UndefinedMethod
```
rondogency: psalm suppress should be below return

tests/Message/ApplePayAuthorizeRequestTest.php
```diff
@@ -0,0 +1,156 @@
...
+        $this->assertSame($request, $request->setApplicationUrl($this->applicationUrl));
+        $this->assertSame($this->applicationUrl, $request->getApplicationUrl());
+    }
+
+    /**
+     * @return void
+     */
+    public function testSendSuccess()
+    {
+        $this->setMockHttpResponse('ApplePayAuthorizeRequestSuccess.txt');
```
rondogency: what is the reason for using txt instead of xml for mock response?

DanayaMel: Because the response that we get back is json.

src/ApplePayGateway.php
DanayaMel: Is there a specific line that you are referring to? The spacing in this doc block is the same as PayPalPurchaseRequest: https://github.com/vimeo/omnipay-vindicia/blob/master/src/Message/PayPalPurchaseRequest.php

DanayaMel:
@rondogency Made all changes, let me know if you have any more suggestions or questions!

src/ApplePayGateway.php
rondogency: just put a double dash between "Element" and "an" to make it easier to read

tests/Message/ApplePayAuthorizeRequestTest.php
rondogency: ok so should we use .json format for mocking response? if we don't need to, can you add a comment on the response format?

src/ApplePayGateway.php
DanayaMel: done!

tests/Message/ApplePayAuthorizeRequestTest.php
DanayaMel: Not necessarily, because only the message is json –– the rest of the response is text. I'll add a comment.

DanayaMel: This is done!

rondogency: ok got it!

src/ApplePayGateway.php
JCManzo: Ahh, my bad. I thought type hints were already part of php 5

PR #43, opened 2018-09-13
Add Autobill fetchInvoice and makePayment request message

rondogency:
This is addressed in v18 upgrade ticket https://github.vimeows.com/Vimeo/vimeo/issues/38027.

We added fetchInvoiceNumbers, fetchInvoice (based on fetchInvoiceNumbers) and makePayment request type to Omnipay, so v18 Autobill payment could go successfully using those requests.


tests/Mock/FetchSubscriptionInvoiceSuccess.xml
```diff
@@ -0,0 +1,50 @@
...
+
+Payments:
+___________________________________________________________________________________
+Total Payments                                                                 0.00
+
+BALANCE                                                                        0.00
+___________________________________________________________________________________
+Current Charges:
+    101                      Recurring Charge, Billing Plan 101    x 1         0.00
+    1217985                  Vimeo Producer                        x 1     [AMOUNT]
```
nickyr: eek please don't put real data on public github! please replace all of this with fake data and force push

tests/Mock/FetchSubscriptionInvoiceSuccess.xml
```diff
@@ -0,0 +1,50 @@
...
+        <returnCode xsi:type="vin:ReturnCode">200</returnCode>
+        <soapId xsi:type="xsd:string">1234567890abcdef1234567890abcdef</soapId>
+        <returnString xsi:type="xsd:string">OK</returnString>
+      </return>
+      <invoice xmlns="" xsi:type="string">
+Vimeo                                      Invoice #: [INVOICE_ID]
+                                                Date: 2020-09-12
+
+Charles Smith
+07030
```
nickyr: no real data plz!

src/Message/AbstractRequest.php
```diff
@@ -421,6 +421,27 @@ public function setProductReference($value)
         return $this->setParameter('productReference', $value);
     }
 
+    /**
+     * Returns the invoice id, which is normally the first element of invoice numbers array
+     *
+     * @return null|string
+     */
+    public function getInvoiceId()
```
nickyr: this is the gateway's identifier, so i think it should be called getInvoiceReference

src/Gateway.php
```diff
@@ -164,6 +164,54 @@ public function updateSubscription(array $parameters = array())
...
 
+    /**
+     * Fetch the invoice number of a subscription
+     *
+     * See Message\FetchSubscriptionInvoiceRequest for more details.
+     *
+     * @param array $parameters
+     * @return \Omnipay\Vindicia\Message\FetchSubscriptionInvoiceRequest
+     */
+    public function fetchSubscriptionInvoiceNumbers(array $parameters = array())
```
nickyr: i think fetchSubscriptionInvoiceReferences would be a better name. it is really weird that vindicia calls it an invoice number in some places and an invoice ID in others and we don't need to repeat that mistake

src/Gateway.php
```diff
@@ -164,6 +164,54 @@ public function updateSubscription(array $parameters = array())
...
+    {
+        /**
+         * @var \Omnipay\Vindicia\Message\FetchSubscriptionInvoiceRequest
+         */
+        return $this->createRequest('\Omnipay\Vindicia\Message\FetchSubscriptionInvoiceRequest', $parameters);
```
nickyr: the way these two functions are reusing the same request object is a bit confusing and it means that you can't enforce the invoice id is passed when it's supposed to be. this means that if you called fetchSubscriptionInvoiceNumbers with the parameters for fetchSubscriptionInvoice, it would just fetch the invoice, when it would actually be more clear to throw an error. can you split it into two separate requests? one can extend the other if you want

src/Message/MakePaymentRequest.php
```diff
@@ -0,0 +1,107 @@
...
+            'action' => $this->getFunction()
+        );
+
+        $data['autobill'] = $subscription;
+        $data['paymentMethod'] = $this->buildPaymentMethod(null);
+        $data['amount'] = $amount;
+        $data['currency'] = null;
+        $data['invoiceId'] = $this->getInvoiceId();
+        $data['overageDisposition'] = null;
+        $data['note'] = 'Payment initiated by makePayment';
```
nickyr: note should be a parameter the user can set, "Payment initiated by makePayment" is a Vimeo-specific note

src/Message/MakePaymentRequest.php
```diff
@@ -0,0 +1,107 @@
...
+ *
+ *   $makePaymentResponse = $gateway->makePayment(array(
+ *       'subscriptionId' => $subscriptionResponse->getSubscriptionId(), // could also do by reference
+ *       'paymentMethodId' => 'cc-123456' // could also do by reference (VID),
+ *       'amount' => 59.95
+ *   ))->send();
+ *
+ *   if ($makePaymentResponse->isSuccessful()) {
+ *       echo 'Payment summary " . $makePaymentResponse->getSummary() . PHP_EOL;';
+ *       vardump($makePaymentResponse->getTransaction());
```
nickyr: var_dump

src/Message/MakePaymentRequest.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+namespace Omnipay\Vindicia\Message;
+
+use stdClass;
+use Omnipay\Common\Exception\InvalidRequestException;
+
+/**
+ * Make payment against an outstanding invoice amount.
```
nickyr: This could use a bit more explanation. It's not clear if the user is being charged.

src/Gateway.php
```diff
@@ -164,6 +164,54 @@ public function updateSubscription(array $parameters = array())
...
+
+    /**
+     * Make payment to a specific amount or an outstanding invoice for a subscription
+     *
+     * See Message\MakePaymentRequest for more details.
+     *
+     * @param array $parameters
+     * @return \Omnipay\Vindicia\Message\MakePaymentRequest
+     */
+    public function makePayment(array $parameters = array())
```
nickyr: here's an idea, and i don't know if it's good: we could call this `charge`. That might make it more clear that a user is actually getting charged. I believe makePayment has the ability to record a payment without actually charging the user too, though. But maybe if we ever supported that, we would call it recordPayment? just trying to think of what naming makes most sense from an Omnipay perspective.

cc @parth21shah

src/Message/MakePaymentRequest.php
```diff
@@ -0,0 +1,107 @@
...
+        );
+
+        $data['autobill'] = $subscription;
+        $data['paymentMethod'] = $this->buildPaymentMethod(null);
+        $data['amount'] = $amount;
+        $data['currency'] = null;
+        $data['invoiceId'] = $this->getInvoiceId();
+        $data['overageDisposition'] = null;
+        $data['note'] = 'Payment initiated by makePayment';
+
```
nickyr: i believe we need to add support for the usePaymentMethodForFutureBilling parameter, which appears to not be documented. but it's in the PHP client library and we use it.

```php
    /**
     * makePayment
     *
     * makes a payment against an AutoBill
     *
     * @param string $srd
     *     sparse return descriptor
     * @param PaymentMethod $paymentMethod
     *     The PaymentMethod used for this Payment.
     * @param decimal $amount
     *     Amount to pay.
     * @param string $currency
     *     ISO 4217 currency code for amount. This must match the currency used for charges on the current invoice (if not specified, the AutoBill/Invoice currency will be used).
     * @param string $invoiceId
     *     ID of Invoice to make payment against. If not specified, the oldest unpaid invoice for this AutoBill will be selected for payment.
     * @param PaymentOverageDisposition $overageDisposition
     *     How to disposition amounts in excess of a required AutoBill payment amount. Defaults to applyToOldestInvoice if not specified
     * @param boolean $usePaymentMethodForFutureBilling
     *     Update the AutoBill to use the included PaymentMethod for future billing. Not Currently Supported
     * @param string $note
     *     Note regarding payment being made
     *
     * @return array - Elements:<ul>
     *         <li>boolean(0/1) '_completed'
     *         <li>int 'returnCode'
     *         <li>string 'returnString'
     *         <li>Update ResponseObject 'data'</ul>
     *
     * @internal UpdateResponse object 'data' contains:<ul>
     * <li>Return 'return' - Basic results of this request (success, failure, error, etc.)
     * <li>Transaction 'transaction' - The transaction generated by the payment attempt. This transaction must be inspected to assess the details of the payment attempt.
     * <li>TransactionAttemptSummary 'summary' - Summary of payment attempt</ul>
     */
    function makePayment($srd, $paymentMethod, $amount, $currency, $invoiceId, $overageDisposition, $usePaymentMethodForFutureBilling, $note) {
        $wsdl = VIN_SOAP_WSDL_PATH . '/AutoBill.wsdl';
        $auth = new Authentication();
        $input = array();
        $input['timeout'] = VIN_SOAP_TIMEOUT;
        $in_arr = array();
        $in_arr['auth'] = $auth;

        $in_arr['autobill'] = $this;
        $in_arr['srd'] = $srd;
        $in_arr['paymentMethod'] = $paymentMethod;
        $in_arr['amount'] = $amount;
        $in_arr['currency'] = $currency;
        $in_arr['invoiceId'] = $invoiceId;
        $in_arr['overageDisposition'] = $overageDisposition;
        $in_arr['usePaymentMethodForFutureBilling'] = $usePaymentMethodForFutureBilling;
        $in_arr['note'] = $note;

        $input['parameters'] = $in_arr;
        $soap = new CallClient();
        $ret = $soap->call($wsdl, 'AutoBill', 'makePayment', $input);
        return $ret;
    }
```

src/Gateway.php
parth21shah: I would go with a charge. I think recordPayment is just for recording purpose it infers payment has ben made we are just recording it

src/Message/MakePaymentRequest.php
parth21shah: I think that's default behavior
we could test that and than add a param maybe?

src/Gateway.php
rondogency: I thought about it for a long time. I think you are right we should split this to be separate request objects. I think ideally we should have individual request object for every method call so that it is easier to make error checking and unit testing. Also those objects should not form inheritance  since they are fetching different things and have no parent-child relationship in oop perspective.

src/Gateway.php
rondogency: I think we should still go with makePayment, since makePayment also indicates the `charge` action to clear an outstanding subscription invoice balance. v18 also has an customer::makePayment methods to pay off balance for an account. I think unless other gateways call it `charge` and you want to unify the naming, I vote for keeping the name as it is.

src/Gateway.php
nickyr: sounds good to me. 

Omnipay does tend to use inheritance for requests that don't really have a parent-child relationship. For example, purchase requests typically extend authorize requests. I'm not sure I agree with the way they do that, but it is definitely the style of the project, so I think it's fine to use inheritance in this way if you want. Either way works.

src/Message/MakePaymentRequest.php
nickyr: could you email Vindicia to confirm that it does nothing and what the default behavior is?

src/Gateway.php
nickyr: i am trying to think of what other omnipay gateways would call it. unfortunately there's not a great way to know... could you take a peek at a few and see if you see anything?

i suspect they wouldn't call it makePayment because "payment" isn't really a word omnipay uses. but then again, idk if other gateways support this at all.

we could do something like makeTransaction... this does generate a transaction record, right?

rondogency: talked with nicky offline and we think it's ok to keep the name as this function is really vindicia specific and doesn't get used in other omnipay libraries.

src/Message/AbstractRequest.php
```diff
@@ -421,6 +421,27 @@ public function setProductReference($value)
         return $this->setParameter('productReference', $value);
     }
 
+    /**
+     * Returns the invoice reference, which is normally the first element of invoice references array
```
nickyr: i think `which is normally the first element of invoice references array` is more confusing than helpful. probably fine without it unless you disagree

src/Message/FetchSubscriptionInvoiceReferencesRequest.php
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Fetch the invoice references of a Vindicia subscription.
+ *
+ * Parameters:
+ * - subscriptionId: Your identifier for the subscription to be fetched. Either subscriptionId
+ * or subscriptionReference is required.
+ * - subscriptionReference: The gateway's identifier for the subscription to be fetched. Either
+ * subscriptionId or subscriptionReference is required.
+ * - invoiceState: the invoice state which limits the returned objects to the specified state:
```
nickyr: is this optional? if so, can you mark it as such?

src/Message/FetchSubscriptionInvoiceRequest.php
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Fetch a Vindicia subscription invoice object.
+ *
+ * Parameters:
+ * - subscriptionId: Your identifier for the subscription to be fetched. Either subscriptionId
+ * or subscriptionReference is required.
+ * - subscriptionReference: The gateway's identifier for the subscription to be fetched. Either
+ * subscriptionId or subscriptionReference is required.
+ * - invoiceReference: the invoice number which uniquely identifies the fetched invoice within subscription
```
nickyr: can you mention that this should be obtained from FetchSubscriptionInvoiceReferencesRequest?

src/Message/FetchSubscriptionInvoiceRequest.php
```diff
@@ -0,0 +1,105 @@
...
+ *   // then take the first invoice reference as invoice identifier to fetch the invoice object
+ *   // normally fetch invoice references only returns an array with one invoice identifier
+ *
+ *.  $invoice_references = $fetchInvoiceReferencesResponse->getInvoiceReferences(); // array of strings
+ *   if (!empty($invoice_references)) {
+ *       $invoice_reference = $invoice_references[0];
+ *       $fetchInvoiceResponse = $gateway->fetchSubscriptionInvoice(array(
+ *           'subscriptionId' => $subscriptionResponse->getSubscriptionId(), // could also do by reference
+ *           'invoiceReference' => $invoice_reference
+ *        ))->send();
```
nickyr: indentation here and below isn't quite lined up with the dollar signs above

src/Message/MakePaymentRequest.php
```diff
@@ -0,0 +1,132 @@
...
+        $subscription->VID = $subscriptionReference;
+
+        $data = array(
+            'action' => $this->getFunction()
+        );
+
+        $data['autobill'] = $subscription;
+        $data['paymentMethod'] = $this->buildPaymentMethod(null);
+        $data['amount'] = $amount;
+        $data['currency'] = null;
```
nickyr: can you add a comment mentioning the default behavior here?

src/Message/Response.php
```diff
@@ -569,6 +569,45 @@ public function getRiskScore()
...
+    public function getInvoiceReferences()
+    {
+        if (isset($this->data->invoicenum)) {
+            return $this->data->invoicenum;
+        }
+        return null;
+    }
+
+    /**
+     * Returns the invoice in PDF/HTML format. For use of FetchSubscriptionInvoice request
```
nickyr: isn't it text format? you're not providing pdf as an option

rondogency:
@nickyr added comments you addressed. I will appreciate if you can make one more approval.

PR #44, opened 2018-09-17
Add deactivate payment method option to create/update payment method request

rondogency:
Add deactivate payment method option to create/update payment method request.

This is used in deactivate_payment_method script in main repo.

Since this is the only place possible to deactivate a payment method, we should enclose the changes only to this specific request.

### DIFF
--- a/src/Message/CreatePaymentMethodRequest.php
+++ b/src/Message/CreatePaymentMethodRequest.php
@@ -30,6 +30,8 @@ use Omnipay\Common\Exception\InvalidRequestException;
  * method is validated. Default is false.
  * - skipCvvValidation: If set to true, CVV validation will not be performed when the payment
  * method is validated. Default is false.
+ * - deactivate: If set to true, the specific payment method will be deactivated on Vindicia side.
+ * Default is false.
  * - updateSubscriptions: If result is true and this request is an update to an existing payment
  * method on an account, Vindicia will update the payment method details on all subscriptions.
  * Default is true.
@@ -142,6 +144,9 @@ class CreatePaymentMethodRequest extends AbstractRequest
         }
         if (!array_key_exists('skipCvvValidation', $parameters)) {
             $parameters['skipCvvValidation'] = false;
+        }
+        if (!array_key_exists('deactivate', $parameters)) {
+            $parameters['deactivate'] = false;
         }
         if (!array_key_exists('updateSubscriptions', $parameters)) {
             $parameters['updateSubscriptions'] = true;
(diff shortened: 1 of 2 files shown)


src/Message/CreatePaymentMethodRequest.php
```diff
@@ -266,6 +271,27 @@ public function setUpdateSubscriptions($value)
...
+    /**
+     * If set to true we set the payment method active flag to be false
+     *
+     * @param bool $value
+     * @return static
+     */
+    public function setDeactivatePaymentMethod($value)
+    {
+        return $this->setParameter('deactivate', $value);
+    }
```
JCManzo: You should probably make the ```deactivate```  string into a const.

src/Message/CreatePaymentMethodRequest.php
```diff
@@ -306,11 +332,16 @@ public function getData($paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
             $this->validate('card');
         }
 
+        $paymentMethod = $this->buildPaymentMethod($paymentMethodType, true);
+        if ($this->getDeactivatePaymentMethod() === true) {
+            $paymentMethod->active = false;
```
JCManzo: Rather than setting ```active``` after creating out the object, could you pass it as an option to ```buildPaymentMethod()```?

rondogency: the buildPaymentMethod() is also reused in other requests, and since this is the only place possible to deactivate a payment method, I choose to enclose the changes only to this specific request.

src/Message/CreatePaymentMethodRequest.php
```diff
@@ -266,6 +271,27 @@ public function setUpdateSubscriptions($value)
         return $this->setParameter('updateSubscriptions', $value);
     }
 
+    /**
+     * If set to false we set the payment method active flag to be false
+     *
+     * @return null|bool
+     */
+    public function getActivatePaymentMethod()
```
nickyr: as discussed offline, parameter function name needs to match the parameter name or Omnipay will fail to initialize it. (see Helper::initialize in Omnipay Common)

src/Message/CreatePaymentMethodRequest.php
```diff
@@ -30,6 +30,8 @@
  * method is validated. Default is false.
  * - skipCvvValidation: If set to true, CVV validation will not be performed when the payment
  * method is validated. Default is false.
+ * - activate: If set to false, the specific payment method will be deactivated on Vindicia side.
```
nickyr: can you update the comment to talk about activation, not deactivation?

src/Message/CreatePaymentMethodRequest.php
```diff
@@ -266,6 +271,27 @@ public function setUpdateSubscriptions($value)
...
     }
 
+    /**
+     * If set to false we set the payment method active flag to be false
+     *
+     * @return null|bool
+     */
+    public function getActivatePaymentMethod()
+    {
+        return $this->getParameter('activate');
```
nickyr: can you make the parameter 'active' instead of 'activate'?

`setActivate(false)` doesn’t really make sense linguistically--it’s not clear that that would deactivate something
whereas `setActive(false)` makes it clear that the payment method will not be active

PR #45, opened 2018-09-17
Fix payment method active flag

rondogency:
The payment method active flag is wrongly named 'activate' previously. So this PR fixes this issue.

### DIFF
--- a/tests/Message/CreatePaymentMethodRequestTest.php
+++ b/tests/Message/CreatePaymentMethodRequestTest.php
@@ -95,14 +95,14 @@ class CreatePaymentMethodRequestTest extends SoapTestCase
     /**
      * @return void
      */
-    public function testActivatePaymentMethod()
+    public function testActive()
     {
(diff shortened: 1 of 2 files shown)


tests/Message/CreatePaymentMethodRequestTest.php
```diff
@@ -245,9 +245,9 @@ public function testGetDataDoNotUpdateSubscriptions()
     /**
      * @return void
      */
-    public function testGetDataDeactivatePaymentMethod()
+    public function testGetDataPaymentMethodNotActive()
     {
-        $data = $this->request->setActivatePaymentMethod(false)->getData();
+        $data = $this->request->setActive(false)->getData();
```
nickyr: in `testGetData`, can you add an assertion that active is true? that'll test the default value

rondogency: I already added that check in testGetData at line 184

nickyr: ah great, i missed that!

PR #46, opened 2018-09-28
Make the return value of response->getInvoiceReferences consistent

rondogency:
Vindicia may mess up the return field if only one invoice num is returned. So we force it to return an array of string.

### DIFF
--- a/src/Message/Response.php
+++ b/src/Message/Response.php
@@ -577,7 +577,7 @@ class Response extends AbstractResponse
     public function getInvoiceReferences()
     {
         if (isset($this->data->invoicenum)) {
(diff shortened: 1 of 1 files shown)


src/Message/Response.php
```diff
@@ -577,6 +577,11 @@ public function getRiskScore()
     public function getInvoiceReferences()
     {
         if (isset($this->data->invoicenum)) {
+            // Vindicia may mess up the field if only one invoice num is returned
```
nickyr: it's actually not really Vindicia's fault, it's the way PHP SOAP parsing works. so probably no need to blame them :-)

rondogency: made the change!

PR #48, opened 2018-10-19
Add support for Avalara tax engine

nickyr:
This PR makes some small changes to better support the Avalara tax engine, which is the standard tax engine that is used with CashBox.

This includes two changes:

- Counts 202 return code as successful: Vindicia will return a 202 if the tax service goes down, but the transaction will still go through, so this should be counted as successful.
- If calculateSalesTax is called without customer, set a dummy value: The Avalara tax engine requires a customer to be set. Vindicia's advice if you don't want to do that is just to set a dummy value.

### DIFF
--- a/src/Message/CalculateSalesTaxRequest.php
+++ b/src/Message/CalculateSalesTaxRequest.php
@@ -84,6 +84,13 @@ class CalculateSalesTaxRequest extends AbstractRequest
             throw new InvalidRequestException('Either the amount or items parameter is required.');
         }
 
+        if ($this->getCustomerId() === null && $this->getCustomerReference() === null) {
+            // With the Avalara tax engine, a customer ID is required for tax to be calculated
+            // Vindicia's advice is to pass a dummy ID if you just want to calculate tax
+            // without a specific user
+            $this->setCustomerId('Dummy ID for tax calculation');
+        }
+
         // skip card validation since we only need the address info
         return array(
             'transaction' => $this->buildTransaction(),
--- a/src/Message/CalculateSalesTaxRequest.php
+++ b/src/Message/CalculateSalesTaxRequest.php
@@ -63,7 +63,7 @@ class CalculateSalesTaxRequest extends AbstractRequest
      * Vindicia's advice is to pass a dummy ID if you just want to calculate tax
      * without a specific user.
      */
-    const DUMMY_CUSTOMER_ID = 'Dummy ID for tax calculation';
+    const DUMMY_CUSTOMER_ID_FOR_TAX_CALCULATION = 'dummy_id_for_tax_calculation';
 
     /**
      * @return string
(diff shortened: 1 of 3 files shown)


src/Message/CalculateSalesTaxRequest.php
```diff
@@ -84,6 +84,13 @@ public function getData()
             throw new InvalidRequestException('Either the amount or items parameter is required.');
         }
 
+        if ($this->getCustomerId() === null && $this->getCustomerReference() === null) {
+            // With the Avalara tax engine, a customer ID is required for tax to be calculated
+            // Vindicia's advice is to pass a dummy ID if you just want to calculate tax
+            // without a specific user
+            $this->setCustomerId('Dummy ID for tax calculation');
```
parth21shah: Should we have some unique ID and make it a constant instead of passing a string?

for example 
```const DUMMY_CUSTOMER_FOR_TAX_CALCULATION = dummy123```

Comment (nickyr) [reply src/Message/CalculateSalesTaxRequest.php]: I think the name should be something ridiculously long so it is unlikely that it ever clashes with a real user for anything. I'm fine with it being a const or not. I didn't bother making one since it only needs to be used in one place and I thought it might be easier to read if it was all kept in one place, but if other people prefer a const too, I'll change it!

File: src/Message/Response.php [hunk 2]
```diff
@@ -12,7 +12,11 @@ class Response extends AbstractResponse
     /**
      * @var array<int>
      */
-    protected static $SUCCESS_CODES = array(200);
+    protected static $SUCCESS_CODES = array(
```
JCManzo: Do we still use the old array syntax here in order to support PHP 5.3 or does that not matter?

nickyr: yup, omnipay supports old phps

src/Message/CalculateSalesTaxRequest.php
DanayaMel: I agree with Parth, I think making it a constant could make it easier for debugging later. It also gives a specific ID to look for instead of searching for a string. I just think it's better practice for the future.

nickyr:
@parth21shah @DanayaMel i moved the dummy customer ID to a const!

(this repo invalidates pr approvals after they become stale, so i'd love a new approval)

src/Message/CalculateSalesTaxRequest.php
```diff
@@ -58,6 +58,13 @@
  */
 class CalculateSalesTaxRequest extends AbstractRequest
 {
+    /**
+     * With the Avalara tax engine, a customer ID is required for tax to be calculated.
+     * Vindicia's advice is to pass a dummy ID if you just want to calculate tax
+     * without a specific user.
+     */
+    const DUMMY_CUSTOMER_ID = 'Dummy ID for tax calculation';
```
parth21shah: @nickyr let us know what you think of changing this?

nickyr: @parth21shah changing it to what?

parth21shah: Something like this const DUMMY_CUSTOMER_FOR_TAX_CALCULATION = dummy123
as mentioned in one of the outdated diff

nickyr: my reply to that was

> I think the name should be something ridiculously long so it is unlikely that it ever clashes with a real user for anything. 

How about I take your name for the const but my value? I'll also replace the spaces with underscores for good measure

parth21shah: Sounds good

nickyr:
alright @parth21shah made the change! could you approve?

PR #50, opened 2019-02-26
Added cacert file and ssl verification to http client.

DanayaMel:
This PR adds a cacert.pem file needed for Guzzle to make successful ssl curl calls. The file was downloaded from here: https://curl.haxx.se/docs/caextract.html. It also adds a `validationUrl` getter to the ApplePayAuthorizeRequest class.


JCManzo:
> It also adds a `validationUrl` getter to the ApplePayAuthorizeRequest class.

I'm not seeing this as a change for this PR

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -358,6 +360,7 @@ protected function createClientRequest($data, array $headers = null)
 
         $config->set('curl.options', $curlOptions);
         $this->httpClient->setConfig($config);
+        $this->httpClient->setDefaultOption('verify', __DIR__ . '/../../certs/cacert.pem');
```
JCManzo: Rather than hard-coding this cert dependency and comiting the cert file manually what do you think about managing this via composer. For example with [composer/ca-bundle](https://github.com/composer/ca-bundle) and then you can use it with Guzzle like so https://github.com/composer/ca-bundle#to-use-with-guzzle

DanayaMel:
> > It also adds a `validationUrl` getter to the ApplePayAuthorizeRequest class.
> 
> I'm not seeing this as a change for this PR. I'm reading this as you added a `validationUrl()` method.

Sorry, I worded that poorly. I meant a validationUrl getter. So the function is `getValidationUrl()` –– found [here](https://github.com/vimeo/omnipay-vindicia/pull/50/files#diff-ba17822789aa8e485b85ea6429963389R327).

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -324,8 +324,9 @@ public function getData()
         $data = array();
 
         // Required request parameters for Apple Session object.
-        $data['merchantIdentifier'] = $this->getMerchantIdentifier();
+        $data['validationUrl'] = $this->getValidationUrl();
```
nickyr: is this related to the cert at all? if not, could you make this change a separate PR? not clear what this is for

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -340,6 +341,7 @@ public function getData()
      * @param array $headers
      *
      * @return static
+     * @psalm-suppress UndefinedMethod
```
nickyr: could you add a comment saying why?

src/Message/ApplePayAuthorizeRequest.php
nickyr: oh that's awesome, good idea @JCManzo!

nickyr: also yay @DanayaMel for figuring out how to do this with guzzle functions!

DanayaMel:
@nickyr @JCManzo changes made

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -340,6 +340,7 @@ public function getData()
      * @param array $headers
      *
      * @return static
+     * @psalm-suppress UndefinedMethod because Guzzle\Client::setDefaultOption() is actually defined.
```
nickyr: lol do you know why it thinks it's not?

PR #52, opened 2019-02-28
Adding an updatePayment method in the apple pay gateway.

DanayaMel:
Adding the update payment method to the Apple Pay gateway.


DanayaMel:
@nickyr Changes made!

PR #57, opened 2019-04-17
Add event dispatcher support

JCManzo:
* Adds the `vimeo/payment-gateway-logger` composer dependency
* Updates the Vindicia gateway to emit Request, Response and Error events that will be listened to by the OmnipayGatewayRequestSubscriber in `vimeo/payment-gateway-logger`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     "require": {
         "vimeo/payment-gateway-logger": "^1.0",
         "omnipay/common": "~2.0",
-        "composer/ca-bundle": "^1.1"
+        "composer/ca-bundle": "^1.1",
+        "ext-soap": "*"
     },
     "require-dev": {
         "omnipay/tests": "~2.0"
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
         }
     },
     "require": {
+        "vimeo/payment-gateway-logger": "^1.0",
         "omnipay/common": "~2.0",
         "composer/ca-bundle": "^1.1"
     },
--- a/src/Message/AbstractRequest.php
+++ b/src/Message/AbstractRequest.php
@@ -110,7 +110,7 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
     const PAYMENT_METHOD_CREDIT_CARD = 'CreditCard';
 
     /**
-     * If chargeback probabilty from risk scoring is greater than this,
+     * If chargeback probability from risk scoring is greater than this,
      * the transaction will fail. By default, every transaction will
      * succeed.
      *
@@ -648,7 +648,7 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
 
     /**
      * Get minimum chargeback probability.
-     * If chargeback probabilty from risk scoring is greater than the
+     * If chargeback probability from risk scoring is greater than the
      * this value, the transaction will fail. If the value is 100,
      * all transactions will succeed.
      *
@@ -661,7 +661,7 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
 
     /**
      * Set minimum chargeback probability.
-     * If chargeback probabilty from risk scoring is greater than the
+     * If chargeback probability from risk scoring is greater than the
      * set value, the transaction will fail. If the value is 100,
      * all transactions will succeed.
      *
--- a/src/Message/AbstractRequest.php
+++ b/src/Message/AbstractRequest.php
@@ -2,6 +2,7 @@
 
 namespace Omnipay\Vindicia\Message;
 
+use Omnipay\Common\Exception\InvalidCreditCardException;
 use Omnipay\Common\Message\ResponseInterface;
 use Omnipay\Vindicia\TestableSoapClient;
 use Omnipay\Vindicia\VindiciaItemBag;
@@ -953,7 +954,6 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
 
             throw $exception;
         }
-
         // reset to how they were before
         ini_set('soap.wsdl_cache_enabled', $originalWsdlCacheEnabled);
         ini_set('soap.wsdl_cache_ttl', $originalWsdlCacheTtl);
@@ -995,7 +995,10 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
      * Set $validateCard to false to skip card validation.
      *
      * @param string $paymentMethodType default null
+     *
      * @return stdClass
+     * @throws InvalidRequestException
+     * @throws InvalidCreditCardException
      */
     protected function buildTransaction($paymentMethodType = null)
     {
@@ -1082,8 +1085,10 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
      *                                        self::PAYMENT_METHOD_APPLE_PAY,
      *                                        null autodetects or sets nothing if no specifying data provided)
      * @param bool $addAttributes default false
+     *
      * @return stdClass
      * @throws InvalidArgumentException if $paymentMethodType is not supported
+     * @throws InvalidCreditCardException
      */
     protected function buildPaymentMethod($paymentMethodType, $addAttributes = false)
     {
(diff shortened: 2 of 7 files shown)


JCManzo:
@nickyr Added unit tests and added events to `ApplePayAuthorizeRequest.php`

src/Message/AbstractRequest.php
```diff
@@ -126,6 +138,7 @@ public function __construct(ClientInterface $httpClient, HttpRequest $httpReques
         parent::__construct($httpClient, $httpRequest);
 
         $this->isUpdate = $isUpdate;
+        $this->eventDispatcher = $httpClient->getEventDispatcher();
```
DanayaMel: Might be helpful to add a description here or line `:41`.

src/Message/AbstractRequest.php
```diff
@@ -966,7 +995,10 @@ protected function isUpdate()
      * Set $validateCard to false to skip card validation.
      *
      * @param string $paymentMethodType default null
+     *
      * @return stdClass
+     * @throws InvalidRequestException
+     * @throws InvalidCreditCardException
```
DanayaMel: What part of this method throws an `InvalidCreditCardException`? Might be helpful to add a comment.

src/TestFramework/TestSubscriber.php
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * The event will be converted to an array before being logged. It will contain the following properties:
+     *     array(
+     *         'request' => \Omnipay\Common\Message\AbstractRequest
+     *     )
+     * @param Event $event
+     * @return void
+     */
+    public function onOmnipayRequestBeforeSend(Event $event)
+    {
```
DanayaMel: Will these tests be written in this PR?

JCManzo: I wasn't planning on doing that for this repo. The logging tests should be done by the client (in our case it'll be in the main vimeo app). I'm just making sure that the Request, Error and Response events are being fired here.

src/Message/AbstractRequest.php
JCManzo: in `buildPaymentMethod`

JCManzo:
@DanayaMel Replied to all of your feedback

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -434,6 +443,9 @@ public function sendData($data)
             );
         // If you try to parse an empty response body, error will be thrown.
         } catch (\RunTimeException $e) {
+            // Log any errors.
+            $eventDispatcher->dispatch(Constants::OMNIPAY_REQUEST_ERROR, new ErrorEvent($e));
```
nickyr: for the regular AuthorizeRequest, it only dispatches an error if it throws an exception. Whereas here, it's going to dispatch an error even though a response object will be returned. that seems inconsistent.

tests/EventEmitterTest.php
```diff
@@ -0,0 +1,155 @@
...
+        ));
+
+        $class = $this;
+        $this->eventDispatcher
+            ->addListener(
+                Constants::OMNIPAY_REQUEST_BEFORE_SEND,
+                /** @return void */
+                function (RequestEvent $event) use ($class) {
+                    $request = $event['request'];
+                    $class->assertInstanceOf('\Omnipay\Vindicia\Message\AuthorizeRequest', $request);
```
nickyr: would this assertion actually fire if no event was dispatched?

tests/EventEmitterTest.php
```diff
@@ -0,0 +1,155 @@
...
+            'EXPIRY_MONTH' => $this->card['expiryMonth'],
+            'EXPIRY_YEAR' => $this->card['expiryYear'],
+            'CVV' => $this->card['cvv'],
+            'PAYMENT_METHOD_ID' => $this->paymentMethodId,
+            'TRANSACTION_ID' => $this->transactionId,
+            'TRANSACTION_REFERENCE' => $this->transactionReference,
+            'RISK_SCORE' => $this->riskScore
+        ));
+
+        $class = $this;
```
nickyr: i think you could just pass $this through, PHP > JS :-)

composer.json
```diff
@@ -24,8 +24,10 @@
         }
     },
     "require": {
+        "vimeo/payment-gateway-logger": "^1.0",
         "omnipay/common": "~2.0",
-        "composer/ca-bundle": "^1.1"
+        "composer/ca-bundle": "^1.1",
+        "ext-soap": "*"
```
nickyr: what's this for?

tests/EventEmitterTest.php
JCManzo: `$this` in closures are only allowed in 5.4 and above unfortunately.

composer.json
JCManzo: It's a safety thing to ensure the PHP soap extension is installed. From here https://getcomposer.org/doc/04-schema.md#package-links

> It is important to list PHP extensions your project requires. Not all PHP installations are created equal: some may miss extensions you may consider as standard (such as ext-mysqli which is not installed by default in Fedora/CentOS minimal installation systems). Failure to list required PHP extensions may lead to a bad user experience: Composer will install your package without any errors but it will then fail at run-time. The composer show --platform command lists all PHP extensions available on your system. You may use it to help you compile the list of extensions you use and require. Alternatively you may use third party tools to analyze your project for the list of extensions used.

`*` is the preferred versioning way of installing due to inconsistent versioning.

tests/EventEmitterTest.php
JCManzo: This callback function is attached to an event listener. It will only be called when a request event is fired otherwise the assertion is never ran.

src/Message/ApplePayAuthorizeRequest.php
JCManzo: yeah good point.

JCManzo: There's no try catch for `$httpRequest->send()` above. I guess I can emit the error event here. @DanayaMel thoughts?

composer.json
nickyr: oh ok, so not actually related to the logging, right?

src/Message/ApplePayAuthorizeRequest.php
nickyr: if send can throw an exception, i think that sounds good

tests/EventEmitterTest.php
nickyr: bummer

tests/EventEmitterTest.php
nickyr: sorry, i'm confused. how do you know an event will always be dispatched?

composer.json
JCManzo: Correct. There are similar issues with curl and json. We need to explicitly tell composer that `ext-curl` and `ext-curl` are being used in this package.

tests/EventEmitterTest.php
JCManzo: Spoke offline; I was misunderstanding. Added unit tests to ensure events were being dispatched.

JCManzo:
@nickyr Added unit tests for ApplePay. Also updated how the location where the Error event is fired for ApplePay.

src/TestFramework/TestSubscriber.php
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * For instance:
+     *
+     *  * array('eventName' => 'methodName')
+     *  * array('eventName' => array('methodName', $priority))
+     *  * array('eventName' => array(array('methodName1', $priority), array('methodName2')))
+     *
+     * @return array The event names to listen to
+     */
+    public static function getSubscribedEvents()
```
nickyr: if you extended https://github.com/vimeo/payment-gateway-logger/blob/master/src/Event/Subscriber/OmnipayGatewayRequestSubscriber.php, you wouldn't need to define this function

tests/EventEmitterTest.php
```diff
@@ -0,0 +1,288 @@
...
+                'currency' => $this->currency,
+                'card' => $this->card,
+                'customerId' => $this->customerId
+            )
+        );
+        $response = $request->send();
+        $this->assertTrue($response->isSuccessful());
+
+        $eventsDispatched = $this->testSubscriber->eventsDispatched;
+        $this->assertEquals(1, $eventsDispatched['omnipay.request.before_send']);
```
nickyr: any reason you're not using the consts for this?

JCManzo: Nope, added!

PR #58, opened 2019-05-02
Update ErrorEvent instantiation to include request object

JCManzo:
* Quick fix to update the `ErrorEvent` constructor to take in the new required request object param.

### DIFF
--- a/src/Message/AbstractRequest.php
+++ b/src/Message/AbstractRequest.php
@@ -951,7 +951,10 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
         } catch (SoapFault $exception) {
             if ($this->eventDispatcher) {
                 // Log any errors.
-                $this->eventDispatcher->dispatch(Constants::OMNIPAY_REQUEST_ERROR, new ErrorEvent($exception));
+                $this->eventDispatcher->dispatch(
+                    Constants::OMNIPAY_REQUEST_ERROR,
+                    new ErrorEvent($exception, $this)
+                );
             }
 
             throw $exception;
--- a/src/Message/ApplePayAuthorizeRequest.php
+++ b/src/Message/ApplePayAuthorizeRequest.php
@@ -425,7 +425,10 @@ class ApplePayAuthorizeRequest extends \Omnipay\Common\Message\AbstractRequest
         } catch (Exception $exception) {
             if ($eventDispatcher) {
                 // Log any errors with the request.
-                $eventDispatcher->dispatch(Constants::OMNIPAY_REQUEST_ERROR, new ErrorEvent($exception));
+                $eventDispatcher->dispatch(
+                    Constants::OMNIPAY_REQUEST_ERROR,
+                    new ErrorEvent($exception, $this)
+                );
             }
 
             throw $exception;
--- a/tests/EventEmitterTest.php
+++ b/tests/EventEmitterTest.php
@@ -117,7 +117,11 @@ class EventEmitterTest extends SoapTestCase
     /**
      * Ensures that 'Request' and 'Error' events are emitted when issuing an improper request.
      *
-     * @psalm-suppress UndefinedMethod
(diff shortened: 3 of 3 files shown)


src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -411,10 +413,14 @@ public function sendData($data)
 
         /** @var EventDispatcherInterface|null $eventDispatcher */
         $eventDispatcher = $this->httpClient->getEventDispatcher();
+        $action = 'startSession';
```
JCManzo: It looks like `$data` differs from the Vindicia AbstractRequest in that it doesn't have an 'action' key so I'm not able to tell which actions this request currently supports. It seems to be just for initiating a session atm? Once we add capture and authorize we can refactor this but for now I've hard-coded the only action that this request does.

nickyr: it's just supposed to log the request name, right? how about just give it a nice request name instead of calling something an "action"? action is a vindicia api specific thing

src/Message/AbstractRequest.php
```diff
@@ -929,7 +929,10 @@ public function sendData($data)
 
         if ($this->eventDispatcher) {
             // Log the request before it is sent.
-            $this->eventDispatcher->dispatch(Constants::OMNIPAY_REQUEST_BEFORE_SEND, new RequestEvent($this));
+            $this->eventDispatcher->dispatch(
+                Constants::OMNIPAY_REQUEST_BEFORE_SEND,
+                new RequestEvent($this, $action)
```
nickyr: this is gonna need more than just $action...some of the fetch requests have the same action (fetchByMerchantId) but are for different objects (autobills, transactions). you could include the object name. or you could include the request name using get_class

JCManzo: Ohhh that's true. Thanks!

JCManzo: Just logging the object name should suffice right? It doesn't seem like logging the action is necessary if it can be used by multiple objects.

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -411,10 +413,14 @@ public function sendData($data)
 
         /** @var EventDispatcherInterface|null $eventDispatcher */
         $eventDispatcher = $this->httpClient->getEventDispatcher();
+        $request_name = 'startSession';
```
nickyr: could you make the name of this one the object name too so it matches the other?

src/Message/AbstractRequest.php
```diff
@@ -927,9 +927,15 @@ public function sendData($data)
         $params = array();
         $params['parameters'] = $data;
 
+        // Parse the name of the class from its namespace
+        $request_name = strtolower(substr(strrchr(get_class($this), "\\"), 1));
```
nickyr: would it be reasonably easy to unit test this?

JCManzo:
The request's name will now come from the Logger that is implemented by the client. This will come in the main Vimeo repo PR. For now, this just passed the request object to the ErrorEvent so that we can access the request that errored when logging.

src/Message/ApplePayAuthorizeRequest.php
```diff
@@ -425,7 +425,10 @@ public function sendData($data)
         } catch (Exception $exception) {
             if ($eventDispatcher) {
                 // Log any errors with the request.
-                $eventDispatcher->dispatch(Constants::OMNIPAY_REQUEST_ERROR, new ErrorEvent($exception));
+                $eventDispatcher->dispatch(
+                    Constants::OMNIPAY_REQUEST_ERROR,
+                    new ErrorEvent($exception, $this)
```
nickyr: hmm might have made more sense for $this to come before $exception, so it lines up with the others. but oh well...

JCManzo: To me it makes sense for it to come after because this is an `ErrorEvent` so the error should come first.

PR #59, opened 2019-05-07
Add support for completeAuthorize and capture for Apple Pay.

DanayaMel:
- Updates `Message\AbstractRequest` for ApplePay completeAuthorize compatibility.
- Adds new `ApplePayCompleteAuthorizeRequest` file.
- Adds tests and mockfile for `ApplePayCompleteAuthorizeRequest`.
- Changed NonstrippingCreditCard's comment for token. 
_(Previous token is actually paymentData that is apart of the Apple Pay token.)_

### DIFF
--- a/src/ApplePayGateway.php
+++ b/src/ApplePayGateway.php
@@ -232,6 +232,7 @@ class ApplePayGateway extends AbstractVindiciaGateway
 
     /**
      * Capture an Apple Pay purchase.
+     * See CaptureRequest for parameter examples.
      *
      * @param array $parameters
      * @return \Omnipay\Vindicia\Message\CaptureRequest
--- a/src/ApplePayGateway.php
+++ b/src/ApplePayGateway.php
@@ -7,9 +7,6 @@ namespace Omnipay\Vindicia;
  * This Apple Pay Gateway provides the functionality to authorize, complete authorize and capture
  * Apple Pay payments for Apple Pay on the web.
  * 
- * // TODO: Add documentation about functionality for capture and completeAuthorize methods to capture
- * // and authorize payments.
- * 
  * Apple Pay is available on all iOS devices with a Secure Element —– an industry-standard, certified
  * chip designed to store payment information safely. On macOS, users must have anApple Pay-capable
  * iPhone or Apple Watch to authorize the payment, or a MacBook Pro with Touch ID.
@@ -35,9 +32,9 @@ namespace Omnipay\Vindicia;
  * (via the response object) you can pass it to the front end to fully load the payment sheet and accept user payment.
  *
  * After the payment sheet is fully loaded, the user can authorize a payment using either Touch or Face ID.
- * The Apple Pay gateway can then use ApplePayCompleteAuthorize to authorize a payment –– no money will be
- * transferred during this step. If the response is successful, the gateway then makes a capture call to 
- * capture a payment and money will be received.
+ * The Apple Pay gateway can then use ApplePayCompleteAuthorize to authorize a payment using the ApplePayPaymentToken
+ * –– no money will be transferred during this step. If the response is successful, the gateway then makes
+ * a capture call to capture a payment and money will be received.
  *
  * <code>
  *    // Setup the gateway with your username and password for Vindicia.
--- a/src/ApplePayGateway.php
+++ b/src/ApplePayGateway.php
@@ -34,6 +34,11 @@ namespace Omnipay\Vindicia;
  * object to retrieve an Apple Pay payment session from Apple's servers. Once the session is retrieved
  * (via the response object) you can pass it to the front end to fully load the payment sheet and accept user payment.
  *
+ * After the payment sheet is fully loaded, the user can authorize a payment using either Touch or Face ID.
+ * The Apple Pay gateway can then use ApplePayCompleteAuthorize to authorize a payment –– no money will be transferred 
+ * during this step. If the response is successful, the gateway then makes a capture call to capture a payment and money
+ * will be received.
+ *
  * <code>
  *    // Setup the gateway with your username and password for Vindicia.
  *    // Include the paths to your Apple Pay Merchant Identity cert, key cert and key cert password.
@@ -79,10 +84,45 @@ namespace Omnipay\Vindicia;
  *    //If successful, the payment sheet should be fully loaded.
  *    $apple_pay_session = $authorizeResponse->getPaymentSessionObject();
  *
- *     // TODO: Add functionality for completeAuthorize() and capture().
  *     // An opaque Apple Pay Session is returned as a response (expires after 5 mins) and it can be sent to
  *     // the front end to fully load the payment sheet. This allows the user to optionally configure their
  *     // payment option and shipping methods (if needed) and submit their payment.
+ *
+ *    //After the user authorizes an Apple Pay payment on the payment sheet using Touch or Face ID, parse the 
+ *    //ApplePayPayment object to retrieve the token. Pass the token to completeAuthorize to authorize a transaction.
+ *    //You may use other fields in the ApplePayPayment object to fill out billing or shipping info.
+ *    $completeAuthorizeResponse = $gateway->completeAuthorize(array(
+ *        //Other params needed to authorize a payment can go here as well.
+ *        'token' => $apple_pay_payment_session_object['token'];
+ *    ))->send();
+ *
+ *    if ($completeAuthorizeResponse->isSuccessful()) {
+ *        // Note: Your transaction ID begins with a prefix you specified in your initial
+ *        // Vindicia configuration. The ID is automatically assigned by Vindicia.
+ *        echo "Transaction id: " . $completeAuthorizeResponse->getTransactionId() . PHP_EOL;
+ *        echo "Transaction reference: " . $completeAuthorizeResponse->getTransactionReference() . PHP_EOL;
+ *        echo "The transaction risk score is: " . $completeAuthorizeResponse->getRiskScore();
+ *    } else {
+ *       // error handling
+ *    }
+ *
+ *    // If the authorization is successful, you may now capture a payment using the transaction ID.
+ *    // Money is transferred during this call.
+ *    $captureResponse = $gateway->capture(array(
+ *        // You can identify the transaction by the transactionId or transactionReference
+ *        // obtained from the authorize response
+ *        'transactionId' => $completeAuthorizeResponse->getTransactionId(),
+ *    ))->send();
+ *
+ *   if ($captureResponse->isSuccessful()) {
+ *        // these are the same as they were on the authorize response, because it is the
+ *        // same transaction
+ *        echo "Transaction id: " . $captureResponse->getTransactionId() . PHP_EOL;
+ *        echo "Transaction reference: " . $captureResponse->getTransactionReference() . PHP_EOL;
+ *    } else {
+ *        // error handling
+ *    }
+ *
  * </code>
  */
 class ApplePayGateway extends AbstractVindiciaGateway
--- a/src/ApplePayGateway.php
+++ b/src/ApplePayGateway.php
@@ -31,11 +31,11 @@ namespace Omnipay\Vindicia;
  * object to retrieve an Apple Pay payment session from Apple's servers. Once the session is retrieved
  * (via the response object) you can pass it to the front end to fully load the payment sheet and accept user payment.
  *
- * After the Apple Pay payment sheet is fully loaded on the frontend. The user can authorize a payment using
+ * After the Apple Pay payment sheet is fully loaded on the frontend, the user can authorize a payment using
  * Touch or Face ID –– this will grant access to the ApplePayPaymentToken. The Apple Pay gateway can authorize 
- * a payment using the ApplePayCompleteAuthorizeRequest with the ApplePayPaymentToken passed in –– no money will 
- * be transferred during this step. If the response is successful, the gateway then makes a capture call using 
- * CaptureRequest to capture a payment and money will be received.
+ * a payment using the ApplePayCompleteAuthorizeRequest with the ApplePayPaymentToken and AuthorizeRequest parameters 
+ * passed in –– no money will be transferred during this step. If the response is successful, the gateway then 
+ * makes a capture call using CaptureRequest to capture a payment and money will be received.
  *
  * <code>
  *    // Setup the gateway with your username and password for Vindicia.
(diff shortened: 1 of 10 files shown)


src/ApplePayGateway.php
```diff
@@ -7,9 +7,6 @@
  * This Apple Pay Gateway provides the functionality to authorize, complete authorize and capture
  * Apple Pay payments for Apple Pay on the web.
  *
```
DanayaMel: Please let me know if the documentation is clear.

src/Message/AbstractRequest.php
```diff
@@ -1139,17 +1139,14 @@ protected function buildPaymentMethod($paymentMethodType, $addAttributes = false
             if (!$this->isUpdate()) {
                 $paymentMethod->type = self::PAYMENT_METHOD_PAYPAL;
             }
-        } elseif ($paymentMethodType === self::PAYMENT_METHOD_APPLE_PAY && $card !== null) {
+        } elseif ($paymentMethodType === self::PAYMENT_METHOD_APPLE_PAY) {
```
DanayaMel: Still testing to be 100% sure, don't believe the card param is necessary.

src/NonStrippingCreditCard.php
```diff
@@ -119,8 +118,7 @@ public function setToken($value)
     }
 
     /**
-     * Gets the token receieved from Apple Pay payment sheet.
-     * Includes the country, zip code, expiration date and account holder name.
+     * Gets the payment token.
```
DanayaMel: This was wrong. The token referred to before was actually the paymentData in the Apple Pay token. I also believe the comment should be more general incase another gateway added uses a token.

DanayaMel: See DataFaker function [here](https://github.com/vimeo/omnipay-vindicia/pull/59/files#diff-0c8d8b8afd3c0bd21b159506e01e7daeR1202).

DanayaMel: Actually, after discussing offline, we're going to have an Apple Pay specific token called `applePayToken`,  since omnipay-common's `token` is typed as a string.

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,92 @@
...
+    public function getApplePayToken()
+    {
+        return $this->getParameter('applePayToken');
+    }
+
+    public function getData($paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
+    {
+        $this->validate('applePayToken');
+
+        return parent::getData(self::PAYMENT_METHOD_APPLE_PAY);
```
JCManzo: This parent function throws `InvalidRequestException` and `InvalidCreditCardException`.  Could you please add this to the doblock for the parent?

src/TestFramework/DataFaker.php
```diff
@@ -1160,13 +1160,24 @@ public function path()
...
+    {
+        return $this->randomCharacters(self::DIGITS . self::ALPHABET_UPPER, 17);
+    }
+
+    /**
+     * The paymentData portion of the token taken from the ApplePayPayment object received from
+     * the Apple Pay payment sheet.
+     * We need to use json encode it so that its parsed as a string instead of an array to
+     *  match token object.
+     *
```
JCManzo: A few typos here .

> The paymentData portion of the token taken from the ApplePayPayment object received from
> \* the Apple Pay payment sheet.

Can you clarify that this is what's returned.

> We need to use json encode it so that its parsed as a string instead of an array to
> \*  match token object.

* Extra space before `match`
* its should be it's
* `use` can be omitted.

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+namespace Omnipay\Vindicia\Message;
+
+/**
+ * After the Apple Pay payment sheet is fully loaded on the frontend. The user can authorize a payment 
+ * using Touch or Face ID –– this will grant access to the ApplePayPaymentToken. You can then authorize a payment
+ *  using this request with the ApplePayPaymentToken as a parameter along with other information needed for
+ *  an authorization (see AuthorizeRequest). No money will be transferred during authorization.
```
JCManzo: A couple of extra leading spaces here.

src/ApplePayGateway.php
```diff
@@ -79,10 +82,45 @@
  *    //If successful, the payment sheet should be fully loaded.
  *    $apple_pay_session = $authorizeResponse->getPaymentSessionObject();
  *
- *     // TODO: Add functionality for completeAuthorize() and capture().
  *     // An opaque Apple Pay Session is returned as a response (expires after 5 mins) and it can be sent to
  *     // the front end to fully load the payment sheet. This allows the user to optionally configure their
  *     // payment option and shipping methods (if needed) and submit their payment.
+ *
+ *    //After the user authorizes an Apple Pay payment on the payment sheet using Touch or Face ID, parse the 
+ *    //ApplePayPayment object to retrieve the token. Pass the token to completeAuthorize to authorize a transaction.
```
nickyr: can you mention that that object comes from the front end?

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,92 @@
...
+ * After authorizing a transaction, call capture to complete the transaction and transfer the money.
+ *
+ * You may use other fields from the ApplePayPayment object to fill out billing info.
+ * This request only requires the 'token' field from the ApplePayPaymentToken extracted from the ApplePayPayment object.
```
nickyr: can you please specify the name of the parameter that this should be passed in as like the other requests do? also could you mention that that object comes from the frontend?

src/ApplePayGateway.php
```diff
@@ -79,10 +82,45 @@
...
+ *    //You may use other fields in the ApplePayPayment object to fill out billing or shipping info.
+ *    $completeAuthorizeResponse = $gateway->completeAuthorize(array(
+ *        //Other params needed to authorize a payment can go here as well.
```
nickyr: can you include them?

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,92 @@
...
+ *           array('name' => 'Item 2', 'sku' => '2', 'price' => '9.99', 'quantity' => 2)
+ *       ),
+ *       'amount' => '23.48', // not necessary since items are provided
+ *       'currency' => 'USD',
+ *       'customerId' => '123456', // will be created if it doesn't already exist
+ *       'card' => array(
+ *           'postcode' => '12345'
+ *       ),
+ *       'paymentMethodId' => 'cc-123456', // this ID will be assigned to the card, which will
+ *        // be attached to the customer's account
```
nickyr: missing some indentation

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,92 @@
...
+     * Gets the ApplePayPaymentToken received from the parsed ApplePayPayment object.
+     *
+     * @return array|null
+     */
+    public function getApplePayToken()
+    {
+        return $this->getParameter('applePayToken');
+    }
+
+    public function getData($paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
```
nickyr: why is credit card the default?

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,92 @@
...
+         */
+        return $this->setParameter('applePayToken', $value);
+    }
+
+    /**
+     * Gets the ApplePayPaymentToken received from the parsed ApplePayPayment object.
+     *
+     * @return array|null
+     */
+    public function getApplePayToken()
```
nickyr: this is already on the parent class

src/NonStrippingCreditCard.php
nickyr: so does this getToken function need to exist? looks unused

DanayaMel: Right, it is. I'll remove all it's uses in this repo.

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,92 @@
...
+class ApplePayCompleteAuthorizeRequest extends AuthorizeRequest
+{
+
+    /**
+     * Sets the ApplePayPaymentToken received from the parsed ApplePayPayment object.
+     *
+     * @param array $value
+     * @return static
+     */
+    public function setApplePayToken($value)
```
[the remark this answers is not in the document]
DanayaMel: In an offline discussion, Nicky and I said that it would be better to leave that information out since Apple Pay's keys could change. Especially when more versions of Apple Pay are added. I think mentioning that this request takes the `ApplePayPaymentToken` may be enough, what do you think?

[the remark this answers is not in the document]
nickyr: that sounds reasonable to me

src/Message/AbstractRequest.php
```diff
@@ -1139,17 +1151,14 @@ protected function buildPaymentMethod($paymentMethodType, $addAttributes = false
...
-            /**
-             * @var \Omnipay\Vindicia\NonStrippingCreditCard $card
-             */
-            $applePay->paymentInstrumentName = $card->getPaymentInstrumentName();
-            $applePay->paymentNetwork = $card->getPaymentNetwork();
-            $applePay->paymentData = $card->getToken();
-            $applePay->transactionReference = $card->getApplePayTransactionReference();
-            $applePay->expirationDate = $card->getExpiryDate(self::VINDICIA_EXPIRATION_DATE_FORMAT);
+            $token = $this->getApplePayToken();
+            $applePay->paymentInstrumentName = $token['paymentMethod']['displayName'];
```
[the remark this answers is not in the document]
DanayaMel: It will never return null because of this line [here](https://github.com/vimeo/omnipay-vindicia/pull/59/files#diff-5e5bf1918fda01730b8630e60acf105eR88), which makes sure that that param is set before continuing.

src/Message/ApplePayCompleteAuthorizeRequest.php
DanayaMel: It does, it's called 'token'. I can change 'field' to 'parameter' to make it more clear.

nickyr: i read this as "take the 'token' field from the ApplePayPayment object and pass it to this request", but that doesn't make it clear that the name of the parameter for this request is _also_ token. the other request objects have lists of parameters that the request takes, starting with the parameter name, which i think is more clear.

src/Message/ApplePayCompleteAuthorizeRequest.php
DanayaMel: Where? Looks okay to me

nickyr: should line up with `// this ID` on the line above i think since it's part of that comment

src/Message/ApplePayCompleteAuthorizeRequest.php
DanayaMel: I can add this line right after this one:
`Pass the extracted 'token' to the 'applePayToken' parameter of this class.`

Look okay to you?

nickyr: yup!

src/Message/AbstractRequest.php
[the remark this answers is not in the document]
JCManzo: Ok could you add a comment explaining the validation that is done before we access `getApplePayToken()` to ensure that it will never be null at this point?

src/Message/ApplePayCompleteAuthorizeRequest.php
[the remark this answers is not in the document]
JCManzo: What about providing a `@link` to https://developer.apple.com/documentation/apple_pay_on_the_web/applepaypaymenttoken?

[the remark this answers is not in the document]
DanayaMel: I'll add a link to the main ApplePayJS API page but I don't think I should get to specific because Apple has been redoing their docs this past month. A lot of pages have been removed.

src/Message/ApplePayCompleteAuthorizeRequest.php
DanayaMel: Fixed this.

DanayaMel: Actually, this needs to stay the way it is. I get this error when I remove it:
```
PHP Fatal error:  Declaration of Omnipay\Vindicia\Message\ApplePayCompleteAuthorizeRequest::getData() must be compatible with Omnipay\Vindicia\Message\AuthorizeRequest::getData($paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
```

DanayaMel:
@nickyr @JCManzo changes made!

src/ApplePayGateway.php
```diff
@@ -75,14 +78,70 @@
...
+ *       'amount' => '23.48', // not necessary since items are provided
+ *       'currency' => 'USD',
+ *       'customerId' => '123456', // will be created if it doesn't already exist
+ *       'card' => array(
+ *          'address1'   => $data->getAddress1(),
+            'address2'   => $data->getAddress2(),
+            'city'       => $data->getCity(),
+            'state'      => $data->getState(),
+            'postcode'   => $data->getPostalCode(),
+            'country'    => $data->getCountry()
```
JCManzo: Syntax error here. You're missing `*` on each of these lines.

src/Message/ApplePayCompleteAuthorizeRequest.php
nickyr: can you make the default apple pay? i suspect psalm is complaining because of the switch from having a parameter to not having a parameter

DanayaMel:
@nickyr @JCManzo changes made!

src/Message/AbstractRequest.php
JCManzo: What's the update on this, was it actually no necessary?

DanayaMel: I was just testing this. When we don't include the card param with a billing address. Vindicia returns an okay (200 'OK'), however, sales tax isn't calculated. We get this error in name values for taxable items:

`"Can't compute sales tax due to lack of shipping and billing addresses"`

So I'll add `&& $card` to this line and update necessary documentation.

nickyr:
looks like this comment from before was missed:

>  could you add to the ApplePayGatewayTest as well?

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,91 @@
+<?php
+
+namespace Omnipay\Vindicia\Message;
+
+/**
+ * After the Apple Pay payment sheet is fully loaded on the frontend. The user can authorize a payment
```
nickyr: period should be a comma

src/Message/AbstractRequest.php
JCManzo: cool 👍

src/Message/ApplePayCompleteAuthorizeRequest.php
```diff
@@ -0,0 +1,91 @@
...
+ *   $gateway->setUsername('your_username');
+ *   $gateway->setPassword('y0ur_p4ssw0rd');
+ *   $gateway->setTestMode(false);
+ *
+ *   // authorize the transaction
+ *   $authorizeResponse = $gateway->completeAuthorize(array(
+ *  *    // ApplePayPaymentToken extracted from ApplePayPayment object.
+ *       // You retrieve the ApplePayPayment object when a user authorizes a payment
+ *       // using the ApplePay payment sheet.
+ *       'applePayToken' => array(
```
nickyr: just doing `'applePayToken' => $apple_pay_token` _might_ be more clear. i'm worried about ppl thinking that they have to parse token and stick it in this array

DanayaMel:
@nickyr changes made!

PR #60, opened 2019-06-12
Modifies getSoapId to return null

DanayaMel:
This PR changes `Response::getSoapId()` to return null instead of throwing an error.

### DIFF
--- a/src/Message/Response.php
+++ b/src/Message/Response.php
@@ -553,7 +553,8 @@ class Response extends AbstractResponse
         if (isset($this->data->return)) {
             return $this->data->return->soapId;
         }
-        throw new InvalidResponseException('Response has no soap id.');
+
+        return null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Message/Response.php
```diff
@@ -553,7 +553,8 @@ public function getSoapId()
         if (isset($this->data->return)) {
             return $this->data->return->soapId;
         }
-        throw new InvalidResponseException('Response has no soap id.');
+
+        return null;
```
JCManzo: Sorry I lied. Could you update the `@return` to account for null?

nickyr:
lgtm! plz release with a minor number increment!

nickyr:
i was too late :-( i think this should have been a minor bump, not a patch, because it could break existing functionality for someone. oh well, it'll be ok

DanayaMel:
@nickyr sry! :/ why am i so bad at version numbers? LOL

nickyr:
lol np. we can discuss beforehand in the future if you want!

JCManzo:
Maybe we can require the expected version bump in the PR title i.e. patch vs minor vs major release. In case anyone has any objections.

DanayaMel:
@JCManzo I agree!

nickyr:
sounds grand!

PR #61, opened 2019-08-12
Allow a subscription cancellation reason to be set and retrieved

nickyr:
Vindicia subscriptions can have a cancellation reason when they get canceled. Sometimes, Vindicia will specify the reason. The merchant can also specify the reason if they make a call to cancel the subscription.

This PR adds support for setting and retrieving the cancel reason to this library.

Vindicia docs about cancel reasons: https://www.vindicia.com/documents/1800APIGuideHTML5/Default.htm#APIGuide/CancelReason_Subobject1.htm%3FTocPath%3DCashBox1800APIGuide%7C4%2520The%2520AutoBill%2520Object%7CAutoBill%2520Subobjects%7C_____4

And here are the possible values: https://www.vindicia.com/documents/2500ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills_with.htm%3FTocPath%3DCashBox2500ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520Canceling%2520AutoBills%7C_____1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ matrix:
         - COMPOSER='composer-psalm.json'
 install: composer install
 script: make $TEST_SUITE
+dist: trusty
 notifications:
   email: false
   slack:
--- a/composer-psalm.json
+++ b/composer-psalm.json
@@ -25,7 +25,9 @@
     },
     "require": {
         "omnipay/common": "~2.0",
-        "vimeo/payment-gateway-logger": "^1.0"
+        "vimeo/payment-gateway-logger": "^1.0",
+        "composer/ca-bundle": "^1.1",
+        "ext-soap": "*"
     },
     "require-dev": {
         "omnipay/tests": "~2.0",
--- a/src/Message/CancelSubscriptionRequest.php
+++ b/src/Message/CancelSubscriptionRequest.php
@@ -15,8 +15,8 @@ use Omnipay\Common\Exception\InvalidRequestException;
  * subscriptionId or subscriptionReference is required.
  * - cancelReason: The reason that the subscription was canceled (optional). Possible values are
  * documented by Vindicia here. Only the reason code needs to be specified:
- * https://www.vindicia.com/documents/2500ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills\
- * _with.htm%3FTocPath%3DCashBox2500ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520\
+ * https://www.vindicia.com/documents/1800ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills\
+ * _with.htm%3FTocPath%3DCashBox1800ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520\
  * Canceling%2520AutoBills%7C_____1
  *
  * Example:
--- a/src/Message/CancelSubscriptionRequest.php
+++ b/src/Message/CancelSubscriptionRequest.php
@@ -14,10 +14,7 @@ use Omnipay\Common\Exception\InvalidRequestException;
  * - subscriptionReference: The gateway's identifier for the subscription to be canceled. Either
  * subscriptionId or subscriptionReference is required.
  * - cancelReason: The reason that the subscription was canceled (optional). Possible values are
- * documented by Vindicia here. Only the reason code needs to be specified:
- * https://www.vindicia.com/documents/1800ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills\
- * _with.htm%3FTocPath%3DCashBox1800ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520\
- * Canceling%2520AutoBills%7C_____1
+ * documented by Vindicia, @see setCancelReason.
  *
  * Example:
  * <code>
@@ -117,6 +114,14 @@ class CancelSubscriptionRequest extends AbstractRequest
 
     /**
      * Set the reason the subscription was canceled
+     *
+     * Possible values:
+     *
+     * https://www.vindicia.com/documents/1800ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills\
+     * _with.htm%3FTocPath%3DCashBox1800ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520 \
+     * Canceling%2520AutoBills%7C_____1
+     *
+     * Only the "reason code" should be specified.
      *
      * @param string $value
      * @return static
--- a/src/Message/CancelSubscriptionRequest.php
+++ b/src/Message/CancelSubscriptionRequest.php
@@ -13,6 +13,11 @@ use Omnipay\Common\Exception\InvalidRequestException;
  * or subscriptionReference is required.
  * - subscriptionReference: The gateway's identifier for the subscription to be canceled. Either
  * subscriptionId or subscriptionReference is required.
+ * - cancelReason: The reason that the subscription was canceled (optional). Possible values are
+ * documented by Vindicia here. Only the reason code needs to be specified:
+ * https://www.vindicia.com/documents/2500ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills\
+ * _with.htm%3FTocPath%3DCashBox2500ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520\
+ * Canceling%2520AutoBills%7C_____1
  *
  * Example:
  * <code>
@@ -101,6 +106,27 @@ use Omnipay\Common\Exception\InvalidRequestException;
 class CancelSubscriptionRequest extends AbstractRequest
 {
     /**
+     * Gets the reason the subscription was canceled
+     *
+     * @return null|string
+     */
+    public function getCancelReason()
+    {
+        return $this->getParameter('cancelReason');
+    }
+
+    /**
+     * Set the reason the subscription was canceled
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setCancelReason($value)
+    {
+        return $this->setParameter('cancelReason', $value);
+    }
+
+    /**
      * The name of the function to be called in Vindicia's API
      *
      * @vreturn string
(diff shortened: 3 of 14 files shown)


.travis.yml
```diff
@@ -25,6 +25,7 @@ matrix:
         - COMPOSER='composer-psalm.json'
 install: composer install
 script: make $TEST_SUITE
+dist: trusty
```
nickyr: Fixes build for php 5.4 and 5.5
 
Travis switched some defaults around. See https://travis-ci.community/t/php-5-4-and-5-5-archives-missing/3723/5

composer-psalm.json
```diff
@@ -25,7 +25,9 @@
     },
     "require": {
         "omnipay/common": "~2.0",
-        "vimeo/payment-gateway-logger": "^1.0"
+        "vimeo/payment-gateway-logger": "^1.0",
+        "composer/ca-bundle": "^1.1",
+        "ext-soap": "*"
```
nickyr: These are in composer.json but were missing here.

This wasn't causing any problems, since the psalm file is used for development only. But better to be consistent.

src/Message/CancelSubscriptionsRequest.php
```diff
@@ -212,7 +238,7 @@ public function getData()
             'action' => $this->getFunction(),
             'force' => true,
             'disentitle' => false,
-            'cancelReasonCode' => null
+            'cancelReasonCode' => $this->getCancelReason()
```
DanayaMel: Wouldn't the reason code get set as the reason code and the description? Should we separate the two?

src/ObjectHelper.php
```diff
@@ -345,6 +345,7 @@ public function buildSubscription(stdClass $object)
             'billingState' => isset($object->billingState) ? $object->billingState : null,
             'startTime' => isset($object->startTimestamp) ? $object->startTimestamp : null,
             'endTime' => isset($object->endTimestamp) ? $object->endTimestamp : null,
+            'cancelReason' => isset($object->cancelReason->reason_code) ? $object->cancelReason->reason_code : null,
```
DanayaMel: Should this be `cancelReasonCode` since only the code is being set?

src/TestFramework/DataFaker.php
```diff
@@ -1111,6 +1111,16 @@ public function riskScore()
         return $this->intBetween(-2, 100);
     }
 
+    /**
+     * Return a subscription cancel reason
+     *
+     * @return string
+     */
+    public function subscriptionCancelReason()
```
DanayaMel: Should this be `subscriptionCancelReasonCode`?

src/Message/CancelSubscriptionsRequest.php
nickyr: No, if you take a look at Vindicia's PHP library, you can only specify the code itself, not the description. Which makes sense, since every code can only have 1 description

src/ObjectHelper.php
nickyr: My thinking here was that the gateway-agnostic thing to do would just be to call it a "cancel reason". Vindicia calls it a "cancel reason code", but that's on them. I figured if multiple gateways were doing this, "cancel reason" would make most sense, and the format of the data might vary wildly by gateway. What do you think? Does that make sense? Could I make it clearer somehow?

src/TestFramework/DataFaker.php
nickyr: see above

src/ObjectHelper.php
DanayaMel: I see your point! "cancel reason" it is!

tests/TestFramework/DataFakerTest.php
```diff
@@ -824,4 +824,14 @@ public function testRiskScore()
...
     }
+
+    /**
+     * @return void
+     */
+    public function testCancelReason()
+    {
+        $cancelReason = $this->faker->subscriptionCancelReason();
+        $this->assertTrue(is_string($cancelReason));
+        $this->assertGreaterThanOrEqual(0, (int) $cancelReason);
```
brendanpeters: Is the reason you're not checking that `$cancelReason < 1000` because we have/could add our own custom-defined codes?

nickyr: yeah. i figured it probably didn't matter. but i'm open to changing it.

brendanpeters: I think it's fine the way it is, I just wanted to make sure I understood

.travis.yml
kaylafuchs: You can remove this from your branch, I had to make the same change a few days ago in order to merge a different PR

src/Message/CancelSubscriptionRequest.php
```diff
@@ -13,6 +13,11 @@
  * or subscriptionReference is required.
  * - subscriptionReference: The gateway's identifier for the subscription to be canceled. Either
  * subscriptionId or subscriptionReference is required.
+ * - cancelReason: The reason that the subscription was canceled (optional). Possible values are
```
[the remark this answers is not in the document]
kaylafuchs: This comment just made me very unsure of which spelling is the normal American one...amazing how your brain just doesn't register things like that 99% of the time 😅

src/Message/CancelSubscriptionRequest.php
```diff
@@ -13,6 +13,11 @@
...
+ * documented by Vindicia here. Only the reason code needs to be specified:
+ * https://www.vindicia.com/documents/2500ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills\
+ * _with.htm%3FTocPath%3DCashBox2500ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520\
+ * Canceling%2520AutoBills%7C_____1
  *
```
kaylafuchs: that url... 🤢

src/Message/CancelSubscriptionRequest.php
```diff
@@ -100,6 +105,27 @@
  */
 class CancelSubscriptionRequest extends AbstractRequest
 {
+    /**
+     * Gets the reason the subscription was canceled
+     *
+     * @return null|string
+     */
+    public function getCancelReason()
+    {
```
kaylafuchs: I've been told a couple times to use type hints instead of docblock comments to specify return values, so that might be nice here (though admittedly inconsistent with the rest of this file)

src/Message/CancelSubscriptionsRequest.php
```diff
@@ -175,6 +180,27 @@ public function setSubscriptionReferences($value)
...
+        return $this->getParameter('cancelReason');
+    }
+
+    /**
+     * Set the reason the subscription was canceled
+     *
+     * @param string $value
+     * @return static
+     */
+    public function setCancelReason($value)
```
kaylafuchs: Same as above re: type hints

tests/Mock/CancelSubscriptionSuccess.xml
```diff
@@ -130,6 +130,10 @@
         </nameValues>
         <subscriptionBalance xmlns="" xsi:type="xsd:decimal">0</subscriptionBalance>
         <statementFormat xmlns="" xsi:type="vin:StatementFormat">DoNotSend</statementFormat>
+        <cancelReason xmlns="" xsi:type="vin:CancelReason">
+          <reason_code xmlns="" xsi:type="xsd:string">[CANCEL_REASON]</reason_code>
```
kaylafuchs: Should probably add this to the mock for CancelSubscriptionsSuccess as well

src/Message/CancelSubscriptionRequest.php
kaylafuchs: Unless we have a different style guide for this repo at the moment and we just want to be consistent!

src/Message/CancelSubscriptionsRequest.php
JCManzo: This repo needs to support PHP 5.3 which doesn't have typehints unfortunately (except for array)

src/Message/CancelSubscriptionRequest.php
JCManzo: See comment here https://github.com/vimeo/omnipay-vindicia/pull/61#discussion_r314784176

src/Message/CancelSubscriptionRequest.php
JCManzo: The link you provide is for v25 of the Cashbox API but the version we actually use is v18

src/Message/CancelSubscriptionsRequest.php
```diff
@@ -22,6 +22,11 @@
  * - subscriptionReferences: Array of references (the gateway's identifiers) of the
  * subscriptions to be canceled. If neither subscriptionIds nor subscriptionReferences are
  * provided, ALL of the user's subscriptions will be canceled.
+ * - cancelReason: The reason that the subscriptions were canceled (optional). Possible values are
+ * documented by Vindicia here. Only the reason code needs to be specified:
+ * https://www.vindicia.com/documents/2500ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills\
+ * _with.htm%3FTocPath%3DCashBox2500ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520\
+ * Canceling%2520AutoBills%7C_____1
```
JCManzo: The link you provide is for v25 of the Cashbox API but the version we actually use is v18

src/Message/Response.php
```diff
@@ -618,6 +618,27 @@ public function getSummary()
         return null;
     }
 
+    /**
+     * Get the reason the subscription was canceled
+     *
+     * One of the reason codes documented here by Vindicia will be returned:
+     * https://www.vindicia.com/documents/2500ProgGuideHTML5/Default.htm#ProgGuide/Canceling_AutoBills\
+     * _with.htm%3FTocPath%3DCashBox2500ProgGuide%7C5%2520Working%2520with%2520AutoBills%7C5.3%2520\
```
JCManzo: The link you provide is for v25 of the Cashbox API but the version we actually use is v18

Also you have this same link in several places and it will need to be updated each time with each version bump. We should find a single place for this link and you can reference it with `@see` or something similar.

tests/Mock/CancelSubscriptionSuccess.xml
nickyr: good thought, but that response doesn't contain the autobills: https://github.com/vimeo/omnipay-vindicia/blob/master/tests/Mock/CancelSubscriptionsSuccess.xml

kaylafuchs: ah yup it does not

nickyr:
Changes made!

PR #62, opened 2019-08-12
Allow custom merchant product ID in success response mock

kaylafuchs:
Allows custom values for `merchantProductId` on the autobill item in the `CreateSubscriptionSuccess` mock response. Intended for use with actual Vimeo product ids, which will enable unit testing certain features that require valid autobill items to be created from a Vindicia response.

### DIFF
--- a/tests/Mock/CreateSubscriptionSuccess.xml
+++ b/tests/Mock/CreateSubscriptionSuccess.xml
@@ -82,7 +82,7 @@
           <index xmlns="" xsi:type="xsd:int">0</index>
           <product xmlns="" xsi:type="vin:Product">
             <VID xmlns="" xsi:type="xsd:string">fabcdef12340987651234098765fabcdef</VID>
-            <merchantProductId xmlns="" xsi:type="xsd:string">[ITEM_MERCHANT_PRODUCT_ID]</merchantProductId>
+            <merchantProductId xmlns="" xsi:type="xsd:string">[PRODUCT_ID]</merchantProductId>
             <status xmlns="" xsi:type="vin:ProductStatus">Active</status>
             <taxClassification xmlns="" xsi:type="xsd:string">TaxExempt</taxClassification>
             <defaultBillingPlan xmlns="" xsi:type="vin:BillingPlan">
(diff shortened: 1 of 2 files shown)


tests/Mock/CreateSubscriptionSuccess.xml
```diff
@@ -82,7 +82,7 @@
           <index xmlns="" xsi:type="xsd:int">0</index>
           <product xmlns="" xsi:type="vin:Product">
             <VID xmlns="" xsi:type="xsd:string">fabcdef12340987651234098765fabcdef</VID>
-            <merchantProductId xmlns="" xsi:type="xsd:string">999999</merchantProductId>
+            <merchantProductId xmlns="" xsi:type="xsd:string">[ITEM_MERCHANT_PRODUCT_ID]</merchantProductId>
```
DanayaMel: I think `MERCHANT_PRODUCT_ID` would suffice to stick with the flow of these xml files. But not a big deal!! Looks good to me!

nickyr: better yet, just PRODUCT_ID? i don't think we use the MERCHANT naming pattern anywhere in these

kaylafuchs: I just wanted to be clear that this sets a property on the nested autobill item so that when the mock is used elsewhere, like in a unit test in a different repo, it's clear what's being set! Most of the properties in the array of custom values you pass to the mock response are top-level properties on the response rather than the autobill item.

nickyr: I think for the purposes of Omnipay, it's actually better not to clarify that. Other gateways might stick the product directly on the subscription, without any "items" concept. And the way you interact with things for those gateways should be the same. 

Not sure we've talked about Omnipay a ton, so hmu if you want to discuss more in person!

kaylafuchs: Yes, would love to! I don't know much about Omnipay at all. Just ended up getting into this repo while working on unit tests for a bug fix.

PR #63, opened 2019-11-27
Fix bug that defaults tax classification on update

lorenzocastillo:
Fixes a bug that would set a default tax classification as TAX_EXEMPT on update.  The default should only be set when creating new products, not on updates.

### DIFF
--- a/src/Message/AbstractRequest.php
+++ b/src/Message/AbstractRequest.php
@@ -144,7 +144,8 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
     }
 
     /**
-     * Initialize request with parameters
+     * Initialize request with parameters.
+     * For new products, the default TAX_CLASSIFICATION is Tax Exempt
      *
      * @param array<string, mixed> $parameters
      * @return static
@@ -154,7 +155,7 @@ abstract class AbstractRequest extends \Omnipay\Common\Message\AbstractRequest
         if (!array_key_exists('timeout', $parameters)) {
             $parameters['timeout'] = self::DEFAULT_TIMEOUT;
         }
-        if (!array_key_exists('taxClassification', $parameters)) {
+        if (!$this->isUpdate() && !array_key_exists('taxClassification', $parameters)) {
             $parameters['taxClassification'] = self::DEFAULT_TAX_CLASSIFICATION;
         }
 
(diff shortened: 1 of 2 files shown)


src/Message/AbstractRequest.php
```diff
@@ -144,7 +144,8 @@ public function __construct(ClientInterface $httpClient, HttpRequest $httpReques
     }
 
     /**
-     * Initialize request with parameters
+     * Initialize request with parameters.
+     * For new products, the default TAX_CLASSIFICATION is Tax Exempt
```
nickyr: any reason tax classification is all capitalized and with an underscore?

lorenzocastillo: I did that because the original variable name is that and I find it usually easier when looking at it outside an IDE doing cmd+F, but no strong preference here.

nickyr: that variable is on Vimeo's private github :-) that should have no influence on this repository--this should be as clear as possible for anyone who may be using it

tests/Message/AbstractRequestTest.php
```diff
@@ -420,6 +420,21 @@ public function testPrices()
         $this->assertSame($prices, $this->request->getPrices());
     }
 
+    public function testDefaultParameterOnCreate() : void
```
jcmanzo: Per the README, this repo supports PHP 5.3+. which does not have support for native return typehints. These should be replaced with a docblock.

tests/Message/AbstractRequestTest.php
```diff
@@ -420,6 +420,21 @@ public function testPrices()
...
+    {
+      $request = Mocker::mock('\Omnipay\Vindicia\Message\AbstractRequest')->makePartial()->shouldAllowMockingProtectedMethods();
+      $request->initialize();
+      $this->assertSame(AbstractRequest::DEFAULT_TAX_CLASSIFICATION, $request->getTaxClassification());
+    }
+
+    public function testDefaultParameterOnUpdate() : void
```
jcmanzo: Same issue here mentioned in https://github.com/vimeo/omnipay-vindicia/pull/63/files#r352650949

tests/Message/AbstractRequestTest.php
nickyr: good catch! that's probably why the build is failing

jcmanzo:
@lorenzocastillo are you familiar with prepping a release for this package once we merge changes? We'll also need to update the dependency version in the main internal Vimeo repo.

PR #64, opened 2019-12-16
Allow disabling purchase AVS/CVN policy on purchase

brendanpeters:
Adds ability to set whether to ignore AVS/CVN policies in `PurchaseRequest`.

### DIFF
--- a/src/Message/PurchaseRequest.php
+++ b/src/Message/PurchaseRequest.php
@@ -66,7 +66,11 @@ class PurchaseRequest extends AuthorizeRequest
         return 'authCapture';
     }
 
-    public function getData(string $paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
+    /**
+     * @param string $paymentMethodType
+     * @return array
+     */
+    public function getData($paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
     {
         $data = parent::getData($paymentMethodType);
 
@@ -86,7 +90,10 @@ class PurchaseRequest extends AuthorizeRequest
         return $this->getParameter('ignoreAvsPolicy');
     }
 
-    public function setIgnoreAvsPolicy(bool $ignore)
+    /**
+     * @param bool $ignore
+     */
+    public function setIgnoreAvsPolicy($ignore)
     {
         $this->setParameter('ignoreAvsPolicy', $ignore);
     }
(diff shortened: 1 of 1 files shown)


src/Message/PurchaseRequest.php
```diff
@@ -62,13 +62,41 @@ protected function getFunction()
...
     }
 
-    public function getData($paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
+    public function getData(string $paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
     {
         $data = parent::getData($paymentMethodType);
 
-        $data['ignoreAvsPolicy'] = false;
-        $data['ignoreCvnPolicy'] = false;
+        $ignore_avs = $this->getIgnoreAvsPolicy();
```
DanayaMel: Could you update the doc block above to include this param, now that it isn't set by default? And a short comment about what this is for?

src/Message/PurchaseRequest.php
```diff
@@ -62,13 +62,41 @@ protected function getFunction()
         return 'authCapture';
     }
 
-    public function getData($paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
+    public function getData(string $paymentMethodType = self::PAYMENT_METHOD_CREDIT_CARD)
```
jcmanzo: This repo supports PHP 5.3+ and scalar type hinting is a PH 7+ feature so this repo needs to rely on PHPdocs

brendanpeters: bummer. :( thanks, I'll change that

brendanpeters:
@DanayaMel updated the docblock
@jcmanzo removed scalar type hints

src/Message/PurchaseRequest.php
```diff
@@ -62,13 +66,51 @@ protected function getFunction()
...
+    /**
+     * @return null|bool
+     */
+    public function getIgnoreAvsPolicy()
+    {
+        return $this->getParameter('ignoreAvsPolicy');
+    }
+
+    /**
+     * @param bool $ignore
```
jcmanzo: Psalm will complain about not having an `@return` here actually

src/Message/PurchaseRequest.php
```diff
@@ -62,13 +66,51 @@ protected function getFunction()
...
+     * @return null|bool
+     */
+    public function getIgnoreCvnPolicy()
+    {
+        return $this->getParameter('ignoreCvnPolicy');
+    }
+
+    /**
+     * @param bool $ignore
+     */
```
jcmanzo: Psalm will complain about not having an @return here

src/Message/PurchaseRequest.php
DanayaMel: I think you have to do:
```
@return static
```

PR #66, opened 2020-11-03
Adds "coerce" param to CancelSubscriptionRequest

DanayaMel:
Adds the option to coerce Vindicia to force cancel a subscription.

Also small syntax fixes in Apple Pay files.

### DIFF
--- a/src/Message/CancelSubscriptionRequest.php
+++ b/src/Message/CancelSubscriptionRequest.php
@@ -85,10 +85,11 @@ use Omnipay\Common\Exception\InvalidRequestException;
  *       // error handling
  *   }
  *
- *   // now we want to cancel the subscription
+ *   // now we want to cancel the subscription, we can choose to coerce the gateway to force cancel the subscription.
  *   $cancelResponse = $gateway->cancelSubscription(array(
  *       // identify the subscription to cancel. could do this by subscriptionReference as well
- *       'subscriptionId' => $subscriptionResponse->getSubscriptionId()
+ *       'subscriptionId' => $subscriptionResponse->getSubscriptionId(),
+ *       'coerce'         => $false // optional: coerce the gateway to force cancel the subscription.
  *   ))->send();
  *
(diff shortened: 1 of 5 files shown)


src/Message/CancelSubscriptionRequest.php
```diff
@@ -160,14 +181,22 @@ public function getData()
...
             'settle' => false,
             'sendCancellationNotice' => false,
-            'cancelReasonCode' => $this->getCancelReason()
+            'cancelReasonCode' => $this->getCancelReason(),
+            'force' => true,
+            'disentitle' => false
         );
+
+        $coerce = $this->getCoercion();
+        if ($coerce) {
```
jcmanzo, line 195: So this is really a shortcut so that we don't need to do the following right?

```PHP
        $response = $gateway->cancelSubscription([
            'subscriptionReference' => $this->vid,
            'force'  => true,
             'disentitle' => true
        ])->send();
```

src/Message/CancelSubscriptionRequest.php
```diff
@@ -112,6 +113,27 @@ public function getCancelReason()
         return $this->getParameter('cancelReason');
     }
 
+    /**
+     * Set the option to force cancel a subscription.
```
jcmanzo, line 117: Could you expand this comment to explain that it would also dis-entitle the user? Not that we use that feature but someone out there someday might

src/Message/CancelSubscriptionRequest.php
DanayaMel: Correct!

src/Message/CancelSubscriptionRequest.php
DanayaMel: Sure!

jcmanzo (approved):

DanayaMel:
JC and I agreed to bump to version v2.21.0 for a minor change

PR #73, opened 2021-09-22
Set product qty

jcmanzo:
We never actually made use of the subscription qty when creating a subscription on Vindicia. This PR makes sure that the quantity is set


jcmanzo (commented):
I can't approve my own PRs but this looks good to me @fredespo

fredespo (approved):

fredespo:
> I can't approve my own PRs but this looks good to me @fredespo

ah right, approved on your behalf 😉

PR #80, opened 2023-03-15
Add PayPalGateway::authorize

fredespo:
This can be used to authorize a PayPal payment method before starting a free trial.

There is a lot of new stuff but is heavily based on existing tests and mock responses for `PayPalPurchaseRequest`.


AdmTal (approved):

kaylafuchs (approved):
✨

PR #84, opened 2023-04-03
Add option to record all test soap calls

fredespo:
This adds the option to record all test SOAP calls made in `TestableSoapClient`. This option is off by default.


kaylafuchs (approved):
Nice, seems useful!

AdmTal (approved):

PR #96, opened 2023-08-07
Avoid setting null payment method type

fredespo:


tests/Message/CreateSubscriptionRequestTest.php
```diff
@@ -316,7 +316,7 @@ public function testGetData(bool $has_items)
             $this->request->setProductReference(null);
             $this->request->setItems($this->items);
         }
-        $data = $this->request->getData(AbstractRequest::PAYMENT_METHOD_CREDIT_CARD);
+        $data = $this->request->getData();
```
fredespo, line 319: @AdmTal changed this test [here](https://github.com/vimeo/omnipay-vindicia/pull/95) to pass in spite of the possibly null payment method type. Now it will pass without Adam's change.

ula (approved):

AdmTal (approved):

PR #104, opened 2024-06-10
Updating all dependencies, including omnipay/common to ^3.0

MelechMizrachi:


composer.json
```diff
@@ -24,12 +24,26 @@
...
-        "omnipay/common": "~2.0",
+        "php": "^7.4|^8.0",
         "composer/ca-bundle": "^1.1",
-        "ext-soap": "*"
+        "ext-soap": "*",
+        "guzzle/common": "^3.9.2",
+        "guzzle/http": "^3.9.2",
+        "guzzle/parser": "^3.9.2",
+        "guzzle/plugin-mock": "^3.9.2",
+        "guzzle/stream": "^3.9.2",
```
MelechMizrachi: These need to go, but for now they can stay. Gives us the ability to do `"symfony/event-dispatcher": ">=2.1",`

MelechMizrachi: Holding this up is https://github.com/vimeo/omnipay-vindicia/blob/update-dependencies/src/Message/AbstractRequest.php#L138 taking in a Guzzle/Http/ClientInterface, instead of the new vindicia/common ClientInterface which also causes this to fail since the signature doesn't match: https://github.com/vimeo/omnipay-vindicia/blob/update-dependencies/src/Message/AbstractRequest.php#L140

MelechMizrachi: Same issue with https://github.com/vimeo/omnipay-vindicia/blob/update-dependencies/src/Message/AbstractHOARequest.php#L28, but just on the __construct level, parent is AbstractRequest so there's no signature clash.

MelechMizrachi: Really we just need this to change: https://github.com/vimeo/omnipay-vindicia/blob/update-dependencies/src/Message/AbstractRequest.php#L145 along with taking an `Omnipay\Common\Http\ClientInterface` instead of a `Guzzle\Http\ClientInterface` and this should work out of the box.

MelechMizrachi: https://github.com/vimeo/omnipay-vindicia/blob/update-dependencies/tests/EventEmitterTest.php#L21 also needs to be updated

